PR #16, opened 2020-06-12
Look for & prioritize "coordinates" attribute when searching.

### LINKED ISSUE #17: use "coordinates" attribute when present
Only search the variables listed in the coordinates attribute when present.

dcherian:
Closes #17

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -119,7 +119,7 @@ def _get_axis_coord(
 
     Returns
     -------
-    str or List[str], Variable name(s) in parent xarray object that matches axis or coordinate `key`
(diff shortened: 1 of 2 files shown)


cf_xarray/accessor.py
```diff
@@ -102,7 +119,7 @@ def _get_axis_coord(var: xr.DataArray, key, error: bool = True, default: Any = N
 
     Returns
     -------
-    str, Variable name in parent xarray object that matches axis or coordinate `key`
+    str or List[str], Variable name(s) in parent xarray object that matches axis or coordinate `key`
```
dcherian, line 122:
```suggestion
    List[str], Variable name(s) in parent xarray object that matches axis or coordinate `key`
```

PR #38, opened 2020-06-19
Set default plotting kwargs

dcherian:
This adds `_plot_decorator` that is used to set default kwargs when appropriate CF attributes are set.

Right now it only looks for `positive="down"` and sets `xincrease, yincrease` appropriately. 

This only works when `x` and `y` are explicitly specified. To work when` x=None` or `y=None` we'll have to do our own autoguessing. This could be useful the y-axis is `Z` or `Y` (priority order) and xaxis is `T`, `X` (priority order). Will leave that for later.

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -252,6 +252,7 @@ def _getattr(
     accessor: "CFAccessor",
     key_mappers: Mapping[str, Mapper],
     wrap_classes: bool = False,
+    extra_decorator: Callable = None,
 ):
     """
     Common getattr functionality.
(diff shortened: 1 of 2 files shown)


cf_xarray/accessor.py
```diff
@@ -261,13 +270,17 @@ def _getattr(
...
+        kwargs based on CF attributes.
     """
-    func = getattr(obj, attr)
+    func: Callable = getattr(obj, attr)
 
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
         arguments = accessor._process_signature(func, args, kwargs, key_mappers)
-        result = func(**arguments)
+        final_func = extra_decorator(func) if extra_decorator else func
```
dcherian, line 282: This is the only design I could think of but it seems a little complicated.

PR #58, opened 2020-07-22
Fix application of multiple mappers.

dcherian:
apply_mapper can now take a tuple of mappers.
It will apply them sequentially. Mappers are expected to return []
if no mapping is possible. A list of results e.g. [[], ["lon"]] is then
unpacked to ["lon"] and returned. If none of the mappers apply or if
more than one mapper can apply, an error is raised.

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -407,7 +407,7 @@ def _getattr(
         newmap = dict()
         unused_keys = set(attribute.keys())
         for key in _AXIS_NAMES + _COORD_NAMES:
-            value = set(apply_mapper(_get_axis_coord, obj, key, error=False,))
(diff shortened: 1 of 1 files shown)


cf_xarray/accessor.py
```diff
@@ -396,7 +407,7 @@ def _getattr(
         newmap = dict()
         unused_keys = set(attribute.keys())
         for key in _AXIS_NAMES + _COORD_NAMES:
-            value = set(apply_mapper(_get_axis_coord, obj, key, error=False))
+            value = set(apply_mapper(_get_axis_coord, obj, key, error=False,))
```
dcherian, line 410:
```suggestion
            value = set(apply_mapper(_get_axis_coord, obj, key, error=False))
```

PR #59, opened 2020-07-23
Adding _get_axis_coord_time_accessor

### LINKED ISSUE #6: Support groupby's datetime accessor behaviour
```
ds.cf.groupby("T.month") → ds.groupby("time.month")
```

jukent:
Closes #6

### DIFF
--- a/cf_xarray/tests/test_accessor.py
+++ b/cf_xarray/tests/test_accessor.py
@@ -306,6 +306,8 @@ def test_dataset_plot(obj):
     ),
 )
 def test_getitem(obj, key, expected_key):
+    assert key in obj.cf
+
     actual = obj.cf[key]
     if isinstance(obj, xr.Dataset):
         expected_key = [expected_key]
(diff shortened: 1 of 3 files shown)


cf_xarray/tests/test_accessor.py
```diff
@@ -119,7 +119,8 @@ def test_rename_like():
     (
         ("resample", {"time": "M"}, {"T": "M"}),
         ("rolling", {"lat": 5}, {"Y": 5}),
-        ("groupby", {"group": "time"}, {"group": "T"}),
+        ("groupby", {"group": "time"}, {"group": "T"}), 
+        ("groupby_time_accessor", {"group": "time.month"}, {"group": "T.month"}),
```
dcherian, line 123:
```suggestion
        ("groupby", {"group": "time.month"}, {"group": "T.month"}),
```

the first element is `attr` which is the name of the function called `getattr(obj, attr)

jukent:
The tests are failing because it is trying to initialize a dataset with a variable "name = 'T.month'" instead of with name = 'T'. We might have to change the test in someway, a check if there is a '.' in the variable name might work.

dcherian:
hmm.. i don't understand that. Also this bot seems too destructive!

PR #66, opened 2020-07-30
Fix mapping for VAR_KEYWORD type kwargs.

dcherian:
This allows plotting with variable names, not jsut cf names


dcherian:
test failure is upstream issue.

PR #75, opened 2020-08-07
install from pypi

keewis:
I recently released `sphinx-autosummary-accessors`, so installing from PyPI should be preferred over installing from github. I'm also planning to create a conda-forge recipe (see xarray-contrib/sphinx-autosummary-accessors#15), but that will probably still take some time.

Edit: also, I noticed the documentation of your `.cf.plot` accessor is broken. You might fix that by using a descriptor other than `property`, see pydata/xarray#3988.


dcherian:
Thanks @keewis !

PR #76, opened 2020-08-17
install sphinx-autosummary-accessors from conda-forge

keewis:
The extension is on `conda-forge`, so you should be able to install that instead of falling back to `pip`.


dcherian:
:) thanks @keewis

PR #81, opened 2020-09-10
fix tests

### LINKED ISSUE #47: fix tests
annoyingly some tests can fail depending on where they are in `test_accessor.py`

For example `test_dicts` fails if it is at the end of the file because `airds.lon.attrs == {}`. 

Sometimes this is an xarray bug (https://github.com/pydata/xarray/issues/4120). 

It would be good to track this down.

jukent:
I found a bug in `test_getitem_errors` which needed a deep copy before deleting the attributes. Other issues have to do with `xr.coarsen` (particularly in `test_kwargs_expand_key_to_multiple_keys`) that was fixed by this PR (https://github.com/pydata/xarray/pull/4360). Once the newest version of Xarray is released these tests will not have to be run in a specific order to pass.

Closes #47


dcherian:
Ah this makes sense. Can you move `test_dicts` to the end of the file and see if tests pass? Or have you done that already? Also you can give yourself credit in `whats-new.rst`

jukent:
> Ah this makes sense. Can you move `test_dicts` to the end of the file and see if tests pass? Or have you done that already? Also you can give yourself credit in `whats-new.rst`

It will not pass unless we use the most current version of Xarray on Github (with the recent PR).

I can add this to whats-new.rst

dcherian:
> It will not pass unless we use the most current version of Xarray on Github (with the recent PR).

The coarsen tests are skipped so that should not be needed?

LEt'add a whats-new note under 0.2.2

jukent:
> The coarsen tests are skipped so that should not be needed?

Hmm that's interesting. If I run "pip install git+git://github.com/pydata/xarray.git" then `test_dicts` passes when I run the tests locally no matter where it is in the file, but if I install the current version of Xarray if fails if it is after `test_kwargs_expand_key_to_multiple_keys` or passes if it is before it. Are you sure that test is skipped?

jukent:
@dcherian I ran `pytest -rxXs` and this was the output:
```
XFAIL cf_xarray/tests/test_accessor.py::test_getitem[area-cell_area-obj0]
  measures not implemented for dataset
XFAIL cf_xarray/tests/test_accessor.py::test_getitem[area-cell_area-obj1]
  measures not implemented for dataset
XFAIL cf_xarray/tests/test_accessor.py::test_getitem[area-cell_area-obj2]
  measures not implemented for dataset
XFAIL cf_xarray/tests/test_accessor.py::test_getitem[area-cell_area-obj3]
  measures not implemented for dataset
SKIPPED [4] cf_xarray/tests/test_accessor.py:120: xarray GH4120. any test after this will fail since attrs are lost
```
So one `coarsen` test is being skipped, but the `test_kwargs_expand_key_to_multiple_keys` also uses `coarsen` and is being run.

dcherian:
ah nice debugging skills!

Let's skip that second coarsen test too then.

cf_xarray/tests/test_accessor.py
```diff
@@ -212,6 +177,7 @@ def test_preserve_unused_keys():
     assert_identical(actual, expected)
 
 
+@pytest.mark.skip(reason="xarray GH4120. any test after this will fail since attrs are lost")
```
dcherian, line 180: Sorry @jukent that was bad advice on my part. Let's just comment out the coarsen lines below and not skip this. We definitely want to check `isel` and `mean` for regressions

doc/whats-new.rst
```diff
@@ -1,5 +1,9 @@
 What's New
 ----------
+v0.2.2 (September 10, 2020)
```
dcherian, line 3:
```suggestion
v0.2.2 (unreleased)
```

cf_xarray/tests/test_accessor.py
dcherian: You could do some advanced skipping using the xarray version but I don't think it's worth it...

dcherian:
Great. Thanks @jukent

Issue #79, opened 2020-08-21
wrap facetgrid objects


dcherian:
This will allow `fg.cf.map_dataarray(x="longitude", ...)`

Adding `xr.plot.FacetGrid` to `_WRAPPED_CLASSES` and some tests should be enough (?)

jukent:
@dcherian I am interested in tackling with this issue, but am not familiar with facetgrid yet. Is this seaborn facetgrid? https://www.tutorialspoint.com/seaborn/seaborn_facet_grid.htm

Is there a line of code that works as `fg.cf.map_dataarray(x="longitude", ...)` is intended to?

dcherian:
Thanks @jukent . This is xarray's facetgrid which  is inspired by seaborn: https://xarray.pydata.org/en/stable/plotting.html#faceting

Here's some example code with output. We want the last commented out line to work. It's pretty similar to groupby and other classes...

``` python
import xarray as xr
import cf_xarray

ds = xr.tutorial.load_dataset("air_temperature")

fg = ds.air.isel(time=slice(6)).cf.plot(col="T", col_wrap=3)  # note .cf.plot; so we can wrap the returned object.
fg.map_dataarray(xr.plot.contour, x="lon", y="lat", colors='w', add_colorbar=False)
# fg.cf.map_dataarray(xr.plot.contour, x="longitude", y="latitude", colors='w', add_colorbar=False)
```

![image](https://user-images.githubusercontent.com/2448579/93623574-77fefc00-f9ce-11ea-987b-df8912d67b69.png)


This may be a little complicated so let me know if you want to chat (i'm a little fuzzy on how this stuff is working right now)

jukent:
Thanks @dcherian! I think what I need to understand the most is the `map_dataarray` function. After some work yesterday (where I discovered that FacetGrid does not work with the most up-to-date version of pandas fyi), I followed through a simple FacetGrid example and found that cf-xarray already worked.

```
airtemps = xr.tutorial.open_dataset("air_temperature")
air = airtemps.air - 273.15
air.attrs = airtemps.air.attrs
air.attrs["units"] = "deg C"

t = air.isel(time=slice(0, 365 * 4, 250))
t.coords
```

where both versions of plotting (with and without cf-xarray)already worked even before adding xr.plot.FacetGrid to the wrapped classes.
`g_simple = t.plot(x="lon", y="lat", col="time", col_wrap=3)`
and
[...]

PR #83, opened 2020-09-17
Facetgrid

jukent:
Address #79


dcherian:
Julia, I think the issue may be the `wrap_classes` arg which is False by default. We want to set it to True for facetgrid plots.

https://github.com/xarray-contrib/cf-xarray/blob/19d80cf48bf99e541f084f97c5ed84b83332d0bd/cf_xarray/accessor.py#L530-L531

dcherian:
```
assert isinstance(rv, xr.plot.FacetGrid)
```
This test should be changed; we aren't returning pure FacetGrid instances anymore.

jukent:
> ```
> assert isinstance(rv, xr.plot.FacetGrid)
> ```
> 
> This test should be changed; we aren't returning pure FacetGrid instances anymore.

That makes sense. I'm not 100% sure what the instances should be yet so I won't change it - at least until I have it working locally.

dcherian:
We can just delete that line and make sure that plotting followed by `map_dataarray` works.

jukent:
> We can just delete that line and make sure that plotting followed by `map_dataarray` works.

There are 3 lines identical to that (on 298, 302, and 312). Should we delete each of these tests?

dcherian:
yes unless you have ideas for something to test.

cf_xarray/tests/test_accessor.py
```diff
@@ -294,24 +294,10 @@ def test_dataarray_plot(obj):
...
-    plt.close()
-
     rv = obj.isel(lat=[0, 1], lon=1).cf.plot.line(x="T", hue="Y")
     assert all([isinstance(line, mpl.lines.Line2D) for line in rv])
     plt.close()
 
-    obj = obj.copy(deep=True)
-    obj.time.attrs.clear()
-    rv = obj.cf.plot(x="X", y="Y", col="time")
-    assert isinstance(rv, xr.plot.FacetGrid)
```
dcherian, line 312: let's only remove the `assert` statements. We still want to try the `.plot` and make sure it doesn't raise an error

jukent: Okay that makes sense. I added the tests back in (without the `assert` statements).

cf_xarray/tests/test_accessor.py
```diff
@@ -295,11 +295,9 @@ def test_dataarray_plot(obj):
     plt.close()
 
     rv = obj.cf.plot(x="X", y="Y", col="T")
```
dcherian, line 297: Now does this work. that will test the changes in this PR

```suggestion
    rv.map_dataarray(xr.plot.contourf, rv, x="X", y="Y", col="T")
```

Issue #82, opened 2020-09-16
enforce dimension ordering?


kthyng:
@dcherian I haven't gotten to integrate cf-xarray into my workflow yet, but, looking ahead, could cf-xarray be a tool for enforcing certain dimension ordering for xarray DataArrays? I find that after calculations sometimes the ordering changes and I haven't figured out a good way to automate dimensional order [time x vertical x y-coord x x-coord]. Sometimes it matters, too. Any thoughts?

dcherian:
the way to do this is `.transpose()`; you can do `.cf.transpose("T", "Z", "Y", "X")` for example. Let me know if it doesn't work for some cases.

kthyng:
Oh right, sorry I should have been specific. Yes, I know transposing works, but you have the know the dimensions ahead of time to call transpose properly. I would hope for something that transposes but given whatever number of the 4 possible dimensions (T, Z, Y, X), puts them in the proper order.

dcherian:
the challenge is that you may pass a Dataset that has an X coordinate with `axis: X` missing (maybe attrs got cleared somehow). Right now that would raise an error, which is good IMO. If cf_Xarray ignored that error then things may silently fail later.

The current solution is
``` python
ds.cf.transpose(*[dim for dim in ["T", "Z", "Y", "X"] if dim in ds.cf.get_valid_keys()])
```

What are you trying to do by enforcing dimension order? Maybe there's something else `cf_xarray` can do elsewhere

kthyng:
That would work if the cf xarray attributes work for a Dataarray for ROMS output. I could have that check as needed in code. I ran across this recently when I did some calculation that changed the order of dimensions in a DataArray and then I used xESMF which assumes typical ordering. So either I would like for my arrays to always be put into proper ordering with a quick command like you listed after anything I run (thinking about xroms here), or all commands to DataArrays should be able to be called by a keyword ordering instead of assuming typical array ordering (in this example that would be a change to xESMF).

dcherian:
> That would work if the cf xarray attributes work for a Dataarray for ROMS output.

if you want `X`, `Y` you may have to add some attrs when you create a dataset in xroms; or change ROMS...

> (in this example that would be a change to xESMF).

This would be desirable. We would like packages that build on xarray to not depend on dimension order. Can you open an issue at https://github.com/pangeo-data/xESMF/issues

kthyng:
[...]

PR #85, opened 2020-09-22
Search through dimension names when coordinates attribute is present

dcherian:
See #82

cc @kthyng I found a bug. Now `ds.temp.cf["X"]` should give `xi_rho` and `ds.temp.cf["latitude"]` should give `lat_rho` (assuming appropriate attributes are set).


kthyng:
awesome! I'll test it out!

kthyng:
This works for me! Thank you very much!

image

Issue #84, opened 2020-09-18
Discovering Axes for ROMS


kthyng:
I am trying to use this excellent package but it doesn't seem to work for ROMS output. What is the best way to use this when the input names are apparently not expected? Here is what I get:

for full Dataset:
![image](https://user-images.githubusercontent.com/3487237/93600262-f2a32980-f984-11ea-8cf4-f768c7203b79.png)

For single DataArray:
![image](https://user-images.githubusercontent.com/3487237/93600280-fa62ce00-f984-11ea-846c-ac7c34b6f6f0.png)

For the T axis, `ocean_time` should be straight-forward since there is only one. For vertical coordinates, there are actually a bunch in my Dataset for all different grid configurations:
![image](https://user-images.githubusercontent.com/3487237/93600407-3138e400-f985-11ea-9dfc-4b9c491b4315.png)

dcherian:
But it's discovered a lot! you can use `'latitude', 'longitude', 'vertical', 'Z'` at this point.

`ds = ds.cf.guess_coord_axis(verbose=True)` will make some guesses (and tell you why) based on regexes that match variable name and `datetime64` / `cftime` objects. This won't add the `axis:X`, `axis:Y`, `axis:Z` attributes for this dataset. But you could do this in `xroms` for example. 

I'm happy to merge more heuristics for `guess_coord_axis` if you have ideas.

dcherian:
Sorry I missed the bit about `z_*` 

I think we could try to mark `z_` variables as `Z` axis by slightly changing this regex if it doesn't work already. Then `guess_coord_axis` will add `axis:Z` to all those variables.

https://github.com/xarray-contrib/cf-xarray/blob/235acac4771597fbd0fe262962450cc1b17ac48c/cf_xarray/accessor.py#L109-L112

Related: There's some discussion about X vs longitude and Y  vs latitude here: https://github.com/xarray-contrib/cf-xarray/issues/23

kthyng:
I didn't mean to imply nothing is working, sorry! It looked to me like the Axes labels are important for what I am trying to do, but I am also still learning what the objectives of cf-xarray are so that I can use it properly. I hate having to specify the grid that I am on when I am doing unambiguous operations to Dataarrays, and it looks like cf-xarray can help me with this, but it also looks like I need the Axes properly labeled for that. I see that having the coordinates pulled out like they already are will allow me to plot without having to specify a different staggered grid for each variable, so that is great.

[...]

PR #86, opened 2020-09-22
Minor fixes to avoid guess attrs for unindexed dim

dcherian:
See #84 

@kthyng this will prevent `guess_coord_axis` from trying set attrs for `xi_*` when there is no corresponding array.


dcherian:
Not sure how to test this without parsing stdout. Let's leave that for later

PR #90, opened 2020-09-23
Separate vertical from Z

### LINKED ISSUE #87: On vertical vs Z
Also I realize that the vertical Axes and Coords are not what I was expecting but my eyes glossed over them earlier.

image

Probably `s_rho` should be the Axes Z (and I am setting this in xroms for this to appear here). But then the associated coord should be `z_rho`, in analogy to the relationship between `xi_rho` and `lon_rho`. Is there an adjustment I can make so that z_rho will become linked as a coord to s_rho in DataArrays?

_Originally posted by @kthyng in https://github.com/xarray-contrib/cf-xarray/issues/82#issuecomment-696975294_

dcherian:
Closes #87 

Z : standard_names, axis
vertical : positive

cc @kthyng @jthielen

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -46,11 +46,18 @@ coordinate_criteria: MutableMapping[str, MutableMapping[str, Tuple]] = {
     "standard_name": {
         "T": ("time",),
         "time": ("time",),
-        "Z": (
+        "vertical": (
             "air_pressure",
             "height",
             "depth",
             "geopotential_height",
+            # computed dimensional coordinate name
+            "altitude",
+            "height_above_geopotential_datum",
+            "height_above_reference_ellipsoid",
+            "height_above_mean_sea_level",
+        ),
+        "Z": (
             "model_level_number",
             "atmosphere_ln_pressure_coordinate",
             "atmosphere_sigma_coordinate",
@@ -63,11 +70,6 @@ coordinate_criteria: MutableMapping[str, MutableMapping[str, Tuple]] = {
             "ocean_s_coordinate_g2",
             "ocean_sigma_z_coordinate",
             "ocean_double_sigma_coordinate",
-            # computed dimensional coordinate name
-            "altitude",
-            "height_above_geopotential_datum",
-            "height_above_reference_ellipsoid",
-            "height_above_mean_sea_level",
         ),
         "latitude": ("latitude",),
         "longitude": ("longitude",),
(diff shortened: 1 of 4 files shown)


cf_xarray/accessor.py
```diff
@@ -82,7 +84,7 @@
     },
     "axis": {"T": ("T",), "Z": ("Z",), "Y": ("Y",), "X": ("X",)},
     "cartesian_axis": {"T": ("T",), "Z": ("Z",), "Y": ("Y",), "X": ("X",)},
-    "positive": {"Z": ("up", "down"), "vertical": ("up", "down")},
```
dcherian, line 85: I don't like this but I don't see an alternative :(

cf_xarray/accessor.py
```diff
@@ -46,11 +46,18 @@
     "standard_name": {
         "T": ("time",),
         "time": ("time",),
-        "Z": (
+        "vertical": (
```
dcherian, line 49: is this a sensible split?

dcherian:
@kthyng much appreciated if you could test this out.

I've been using
``` python
ds = xr.Dataset()
ds["s_rho"] = ("s_rho", np.linspace(-1, 0, 10), {"standard_name": "ocean_s_coordinate_g2"})
ds.coords["z_rho"] = ("s_rho", np.linspace(-100, 0, 10), {"positive": "up"})
ds["temp"] = ("s_rho", np.linspace(20, 30, 10), {"coordinates": "z_rho"})
```

kthyng:
I can get it so that z_rho is recognized as coordinates for the whole dataset, but not for the temp DataArray in particular. I do set 
`ds['temp'].attrs['coordinates'] = 'lon_rho lat_rho z_rho ocean_time'`

dcherian:
OK there was bug if the coordinates string refers to a variable name that isn't carried around by the DataArray. Is `z_rho in ds.temp.coords` True?

kthyng:
> OK there was bug if the coordinates string refers to a variable name that isn't carried around by the DataArray. Is `"z_rho" in ds.temp.coords` True?

Yes.

> It would be useful to see reprs for `ds` `ds.temp` with expanded attrs for temp and `z_rho` and `s_rho`.

`ds.temp`:

image

`ds.z_rho`:

image

`ds.s_rho`:

image


> Oh you should delete `positive` attribute on `s_rho`.

I did do that.

dcherian:
hmmm i don't see what's wrong; the example code above mirrors what you have.

Is it raising an error because `ds.temp.cf["vertical"]` finds both `s_rho` and `z_rho` or is it just not finding anything?

kthyng:
`ds.temp.cf["vertical"]` is not finding anything.

image

Does `z_rho` need more attributes? Oh, but it is found by the full dataset (haven't touched `s_w` yet):

image

dcherian:
Merging so that we can iterate more conveniently. Please open issues if you find anything.

PR #95, opened 2020-09-25
Fix bug when coordinates is in both attrs and encoding

### LINKED ISSUE #94: associating vertical coordinates with dataarrays in dataset in ROMS
I have updated to the newest stuff in github (so includes #90) but have the same problem as before: cf-xarray can find everything for Axes and Coordinates in the general Dataset, but I can't connect the z coordinates properly to individual DataArrays within the Dataset, as shown here with `temp` (no "vertical" coords). I have tried with and without adding the line:
`ds['temp'].attrs['coordinates'] = 'lon_rho lat_rho z_rho ocean_time'` to test with temp in particular but there is no difference in describe.

image

dcherian:
Closes #94


dcherian:
image

@kthyng the opendap dataset was very useful!

PR #102, opened 2020-10-08
added time frequencies to regex for T

kthyng:
My goal in this is to expand what is matched to "count" as time when `guess_coord_axis` is run to include other time frequencies that might result from for example a groupby call. With the change in this PR, the following is now possible

image

This avoids needing to add attributes after the calculation 

`ds.temp.isel(xi_rho=300,s_rho=-1).cf.groupby("T.hour").mean(skipna=False)`

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -109,7 +109,7 @@ coordinate_criteria["long_name"] = coordinate_criteria["standard_name"]
 
 #: regular expressions for guess_coord_axis
 regex = {
-    "time": "time[0-9]*",
+    "time": "time[0-9]*|min|hour|day|week|month|year",
     "vertical": (
         "(lv_|bottom_top|sigma|h(ei)?ght|altitude|depth|isobaric|pres|"
         "isotherm)[a-z_]*[0-9]*"


dcherian:
Thanks! Any interest in adding a test? We don't test these regexes or coordinate criteria yet so it is not a big deal.

cf_xarray/accessor.py
```diff
@@ -109,7 +109,7 @@
 
 #: regular expressions for guess_coord_axis
 regex = {
-    "time": "time[0-9]*",
+    "time": "time[0-9]*|min|hour|day|week|month|year",
```
dcherian, line 112: Can you add the attributes listed here: https://xarray.pydata.org/en/stable/generated/xarray.core.accessor_dt.DatetimeAccessor.html and https://xarray.pydata.org/en/stable/generated/xarray.core.accessor_dt.TimedeltaAccessor.html? (the ones that make sense anyway like `dayofyear`)

kthyng:
I hope to get back to this this week.

dcherian:
Sounds good. We could also detect `ocean_time` or any `*time*` variable.

dcherian:
Merging. We can upgrade this functionality later.

PR #104, opened 2020-10-13
Adding functions to find Axis names and Coord names instead of both together

kthyng:
Since Axis names map to xarray dimensions and Coordinate names map to xarray coordinates, I'd like to be able to return a list of each for a given DataArray separate. So, I introduced functions `get_valid_axis` and `get_valid_coords` which are smaller versions of `get_valid_keys` to do this. I am certainly not tied to the names in fact I am bothered by one having an "s" and not the other but wasn't sure what to do to be consistent.

I also wasn't being careful in my original lack of branching so I had to remove a change that is in #102 to keep things separate.  Wasn't sure the best way to do that. Also realize I probably need to do more things like tests but thought I should see what you think of this.

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1328,6 +1328,91 @@ class CFDatasetAccessor(CFAccessor):
 
         return self._maybe_to_dataarray(obj)
 
+    def decode_vertical_coords(self, prefix="z"):
+        """
+        Decode parameterized vertical coordinates in place.
+
+        Parameters
+        ----------
+        prefix: str, optional
+            Prefix for newly created z variables.
+            E.g. ``s_rho`` becomes ``z_rho``
+
+        Returns
+        -------
+        None
+
+        Notes
+        -----
+
+        Will only decode when the ``formula_terms`` and ``standard_name`` attributes
+        are set on the parameter (e.g ``s_rho`` )
+
+        Currently only supports ``ocean_s_coordinate_g1`` and ``ocean_s_coordinate_g2``.
+
+        .. warning::
+           Very lightly tested. Please double check the results.
+        """
+        import re
+
+        ds = self._obj
+        dims = _get_axis_coord(ds, "Z")
+
+        requirements = {
+            "ocean_s_coordinate_g1": {"depth_c", "depth", "s", "C", "eta"},
+            "ocean_s_coordinate_g2": {"depth_c", "depth", "s", "C", "eta"},
+        }
+
+        for dim in dims:
+            suffix = dim.split("_")
+            zname = f"{prefix}_" + "_".join(suffix[1:])
+
+            if (
+                "formula_terms" not in ds[dim].attrs
+                or "standard_name" not in ds[dim].attrs
+            ):
+                continue
+
+            formula_terms = ds[dim].attrs["formula_terms"]
+            stdname = ds[dim].attrs["standard_name"]
+
+            # map "standard" formula term names to actual variable names
+            terms = {}
+            for mapping in re.sub(": ", ":", formula_terms).split(" "):
+                key, value = mapping.split(":")
+                if value not in ds:
+                    raise KeyError(
+                        f"Variable {value!r} is required to decode coordinate for {dim} but it is absent in the Dataset."
+                    )
+                terms[key] = ds[value]
+
+            absent_terms = requirements[stdname] - set(terms)
+            if absent_terms:
+                raise KeyError(f"Required terms {absent_terms} absent in dataset.")
+
+            if stdname == "ocean_s_coordinate_g1":
+                # S(k,j,i) = depth_c * s(k) + (depth(j,i) - depth_c) * C(k)
+                S = (
+                    terms["depth_c"] * terms["s"]
+                    + (terms["depth"] - terms["depth_c"]) * terms["C"]
+                )
+                # z(n,k,j,i) = S(k,j,i) + eta(n,j,i) * (1 + S(k,j,i) / depth(j,i))
+                ds.coords[zname] = S + terms["eta"] * (1 + S / terms["depth"])
+
+            elif stdname == "ocean_s_coordinate_g2":
+                # make sure all necessary terms are present in terms
+                # (depth_c * s(k) + depth(j,i) * C(k)) / (depth_c + depth(j,i))
+                S = (terms["depth_c"] * terms["s"] + terms["depth"] * terms["C"]) / (
+                    terms["depth_c"] + terms["depth"]
+                )
+                # z(n,k,j,i) = eta(n,j,i) + (eta(n,j,i) + depth(j,i)) * S(k,j,i)
+                ds.coords[zname] = terms["eta"] + (terms["eta"] + terms["depth"]) * S
+
+            else:
+                raise NotImplementedError(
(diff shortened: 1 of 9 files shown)


dcherian:
Thanks. This is a nice feature. I'm thinking we want to
1. Change `.get_valid_keys()` to `.cf.keys()` (matching `dict.keys()`)
2. Make these functions `.cf.axis_keys()`, `.cf.coord_keys()`.

How does that sound?

kthyng:
> 1. Change `.get_valid_keys()` to `.cf.keys()` (matching `dict.keys()`)
> 2. Make these functions `.cf.axis_keys()`, `.cf.coord_keys()`.
> 
> How does that sound?

I like both. Also would be fine with `cf.axes` and `cf.coords`.

kthyng:
Yes I like your proposal. 

What is the next step?

kthyng:
Ok is `.cf.standard_names` the same as the existing `cf.get_standard_names`? I'll try for (1) but will need corrections.

cf_xarray/accessor.py
```diff
@@ -897,7 +906,7 @@ def get_valid_axis(self) -> Set[str]:
 
         return set(varnames)
 
-    def get_valid_coords(self) -> Set[str]:
+    def coordinates(self) -> Set[str]:
```
dcherian, line 908:
```suggestion
    @property
    def coordinates(self) -> Set[str]:
```

Let's add `@property` for all of these (except `keys`) so we can do `.cf.coordinates` instead of `.cf.coordinates()`

cf_xarray/accessor.py
```diff
@@ -850,6 +850,15 @@ def describe(self):
         print(text)
 
     def get_valid_keys(self) -> Set[str]:
+        
+        warnings.warn(
+                      """Now called `keys` and `get_valid_keys` will 
+                      be removed in a future version.""",
+                      DeprecationWarning,
+                      )
```
dcherian:
```suggestion
                      )
                      
        return self.keys()
```

for backward compatibility

cf_xarray/accessor.py
```diff
@@ -917,7 +926,35 @@ def get_valid_coords(self) -> Set[str]:
...
+
+        return set(varnames)
+
     def get_standard_names(self) -> List[str]:
+        
+        warnings.warn(
+                      """Now called `standard_names` and `get_standard_names` 
+                      will be removed in a future version.""",
+                      DeprecationWarning,
+                      )
```
dcherian:
```suggestion
                      )
        return self.standard_names
```

dcherian (commented):
Can we also add a `whats-new.rst` entry, add to `api.rst`, and update the calls in the example notebook. We should also add a section demonstrating the new properties in that notebook (it can be brief, just show `.axes` and `.coordinates` and mention that `.cell_measures` and `.standard_names` are also available).

dcherian:
We should also add some tests with one of the example datasets.

cf_xarray/accessor.py
kthyng: Ok. I haven't done interactive PR stuff before but I will commit this and then add `@property` to the other new functions except `keys`.

cf_xarray/accessor.py
kthyng: Oops, forgot that part.

kthyng: did this in a separate commit since already changed the repo.

kthyng:
> Can we also add a `whats-new.rst` entry, add to `api.rst`, and update the calls in the example notebook. We should also add a section demonstrating the new properties in that notebook (it can be brief, just show `.axes` and `.coordinates` and mention that `.cell_measures` and `.standard_names` are also available).

> We should also add some tests with one of the example datasets.

I will try to do these things.

kthyng:
- [ ] add a whats-new.rst entry
- [ ] add to api.rst
- [ ] update the calls in the example notebook
- [ ] add a section demonstrating the new properties in that notebook (it can be brief, just show .axes and .coordinates and mention that .cell_measures and .standard_names are also available).
- [ ] add some tests with one of the example datasets.

dcherian:
The `flake8` failure is 
```
 flake8...................................................................Failed
- hook id: flake8
- exit code: 1

cf_xarray/accessor.py:855:47: W291 trailing whitespace
cf_xarray/accessor.py:896:75: W291 trailing whitespace
cf_xarray/accessor.py:917:75: W291 trailing whitespace
cf_xarray/accessor.py:938:75: W291 trailing whitespace
cf_xarray/accessor.py:956:56: W291 trailing whitespace
```

cf_xarray/accessor.py
```diff
@@ -850,6 +850,16 @@ def describe(self):
         print(text)
 
     def get_valid_keys(self) -> Set[str]:
+
+        warnings.warn(
+            """Now called `keys` and `get_valid_keys` will
```
dcherian, line 855:
```suggestion
            "Now called `keys` and `get_valid_keys` will
```

cf_xarray/accessor.py
```diff
@@ -850,6 +850,16 @@ def describe(self):
...
+                      be removed in a future version.""",
```
dcherian, line 856:
```suggestion
                      be removed in a future version.",
```

cf_xarray/accessor.py
```diff
@@ -877,7 +887,81 @@ def get_valid_keys(self) -> Set[str]:
         varnames.extend(self.get_standard_names())
         return set(varnames)
 
+    @property
+    def axes(self) -> Set[str]:
+        """
+        Utility function that returns valid Axis names for .cf[].
```
dcherian, line 893:
```suggestion
        Property that returns valid Axis names for ``.cf[]``.
```

cf_xarray/accessor.py
```diff
@@ -877,7 +887,81 @@ def get_valid_keys(self) -> Set[str]:
...
+
+        This is useful for checking whether a key is valid for indexing, i.e.
+        that the attributes necessary to allow indexing by that key exist.
+        However, it will only return the Axis names, not Coordinate names.
+
+        Returns
+        -------
+        Set of valid Axis names that can be used with __getitem__ or .cf[key].
```
dcherian, line 901:
```suggestion
        Set of valid Axis names that can be used with ``__getitem__`` or ``.cf[key]``.
```

cf_xarray/accessor.py
```diff
@@ -877,7 +887,81 @@ def get_valid_keys(self) -> Set[str]:
...
+        Utility function that returns valid Coordinate names for .cf[].
+
+        This is useful for checking whether a key is valid for indexing, i.e.
+        that the attributes necessary to allow indexing by that key exist.
+        However, it will only return the Coordinate names, not Axis names.
+
+        Returns
+        -------
+        Set of valid Coordinate names that can be used with __getitem__ or .cf[key].
+        """
```
dcherian, line 923: Similarly here.

cf_xarray/accessor.py
kthyng: I changed this separately and it is on one line now.

cf_xarray/accessor.py
kthyng: Same, changed separately and now on one line.

dcherian:
Thanks @kthyng!

kthyng:
Thanks for your help along the way @dcherian!

PR #112, opened 2020-11-16
Fix cell measures parser

### LINKED ISSUE #111: Multiple cell measures error
## Problem
If I'm understanding it right, I should be able to assign multiple `cell_measures`.
The attribute format should be `key0: value0 key1: value1 ....`, but this is currently raising an error whenever I use `cf_xarray` to select variables:
https://github.com/xarray-contrib/cf-xarray/blob/4249cf1986a8e6115deeaedb51e947731aa66767/cf_xarray/accessor.py#L369-L372
```
ValueError: attrs['cell_measures'] = 'key0: value0 key1: value1' is malformed.
```

## Fix
I think just changing line 370 would fix it (I didn't check if it breaks anything else yet, I'm happy to send a PR if this is a bug).
```python
attr = 'key0: value0 key1: value1'
current_strings = [s.strip() for s in attr.strip().split(":")]
print('Current:', current_strings)

proposed_strings = [s1 for s0 in attr.strip().split(":") for s1 in s0.split()]
print('Proposed:', proposed_strings)
```
```
Current: ['key0', 'value0 key1', 'value1']
Proposed: ['key0', 'value0', 'key1', 'value1']
```

malmans2:
Closes #111

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -367,7 +367,7 @@ def _get_measure(da: Union[DataArray, Dataset], key: str) -> List[str]:
         )
 
     attr = da.attrs["cell_measures"]
-    strings = [s.strip() for s in attr.strip().split(":")]
+    strings = [s for scolons in attr.split(":") for s in scolons.split()]
     if len(strings) % 2 != 0:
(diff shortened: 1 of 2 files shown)


dcherian:
Thanks @malmans2

Issue #99, opened 2020-10-05
DataArray coordinates are dropped


malmans2:
It looks like when I pull out a DataArray from a Dataset using `cf_xarray` the coordinates are dropped (while `xarray` keeps them).
Here is an example:

```python
import xarray as xr
import cf_xarray
import numpy as np
```

```python
x = 10
y = 20
ds = xr.Dataset(
    dict(
        lon=xr.DataArray(
            np.arange(x), dims=("x"), attrs=dict(standard_name="longitude"),
        ),
        lat=xr.DataArray(
            np.arange(y), dims=("y"), attrs=dict(standard_name="latitude"),
        ),
        var=xr.DataArray(
            np.random.rand(x, y),
            dims=("x", "y"),
            attrs=dict(standard_name="cf_var_name"),
        ),
    )
)
ds = ds.set_coords(["lon", "lat"])
```

```python
ds.cf.describe()
```

    Axes:
    	X: []
    	Y: []
    	Z: []
    	T: []
    
    Coordinates:
    	longitude: ['lon']
    	latitude: ['lat']
    	vertical: []
    	time: []
    
    Cell Measures:
    	area: unsupported
    	volume: unsupported
    
    Standard Names:
    	cf_var_name: ['var']
    

```python
print(ds["var"].coords)
```

    Coordinates:
        lon      (x) int64 0 1 2 3 4 5 6 7 8 9
        lat      (y) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19

```python
print(ds.cf["cf_var_name"].coords)
```

    Coordinates:
        *empty*

```python
# OK if I pull out the DataArray directly from ds
ds["var"].cf.plot(x="longitude", y="latitude")
```

    <matplotlib.collections.QuadMesh at 0x7fbf13acf690>

![image](https://user-images.githubusercontent.com/22245117/95093863-24f0ad00-0721-11eb-9668-48a6db80fa2c.png)

```python
# Problem if I pull out the DataArray from ds.cf
ds.cf["cf_var_name"].cf.plot(x="longitude", y="latitude")
```

    ---------------------------------------------------------------------------

    KeyError                                  Traceback (most recent call last)

    /noc/msm/scratch/climate/malmans/miniconda3/envs/overflows/lib/python3.7/site-packages/xarray/core/dataarray.py in _getitem_coord(self, key)
        628         try:
    --> 629             var = self._coords[key]
        630         except KeyError:

[...]

PR #115, opened 2020-11-16
Allow any cell measure

malmans2:
See https://github.com/xarray-contrib/cf-xarray/issues/99#issuecomment-727984869

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -345,7 +345,7 @@ def _get_measure(da: Union[DataArray, Dataset], key: str) -> List[str]:
     ----------
     da: DataArray
         DataArray belonging to the coordinate to be checked
-    key: str
+    key: str, ["area", "volume"]
         key to check for.
     error: bool
         raise errors when key is not found or interpretable. Use False and provide default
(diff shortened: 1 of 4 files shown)


malmans2:
Not sure what's up with the pre-commit action. It's running fine on my machine.

malmans2 (commented):
@dcherian I've commented out this check so `get_associated_variable_names` can use `_get_measure`. In the `_get_measure`  docstring there are two arguments that I don't think are actually used: `error` and `default`. Do you want to implement those and use `error=False` in  `get_associated_variable_names` , or we can just get rid of the `valid_keys` check?

malmans2:
Here is the part I've commented out:
https://github.com/xarray-contrib/cf-xarray/blob/86f3e511fbef9c70158afc767886f52ee5113d2d/cf_xarray/accessor.py#L339-L370

malmans2:
Actually, I realized I don't have to use `_get_measure` in `get_associated_variable_names`.
Sorry, this was a bit messy, but I think it's good to go now.

dcherian:
Thanks @malmans2 . I think we could allow indexing `.cf.cell_measures` by any arbitrary cell measure name, not just area, volume. WDYT? 

We could also show these extra measures in `.cf.describe()`

dcherian:
Actually this will take some effort. This LGTM for now.

PR #118, opened 2020-11-30
new cell measures implementation

### LINKED ISSUE #116: Show and allow indexing by extra cell measures
As follow on from #115, it would be good to show arbitrary cell measures in `.cf.describe` and `.cf.cell_measures`. I think this would take some effort.

malmans2:
Closes #116 

After some thought I came up with this implementation. 
I kept the hardcoded `_CELL_MEASURES`, so defined CF measures (area and volume) are always displayed by `describe()`.

User defined measures are added to `_CELL_MEASURES` when the `CFAccessor` is initialized.

I had to comment out `_get_measure_variable` in the `_build_docstring` (not sure if it needs to be substituted with something else).

What do you think?

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -338,26 +338,20 @@ def _get_measure_variable(
 
 def _get_measure(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     """
-    Translate from cell measures ("area" or "volume") to appropriate variable name.
+    Translate from cell measures to appropriate variable name.
     This function interprets the ``cell_measures`` attribute on DataArrays.
 
     Parameters
     ----------
     obj: DataArray, Dataset
         DataArray belonging to the coordinate to be checked
-    key: str, ["area", "volume"]
+    key: str
         key to check for.
 
     Returns
     -------
     List[str], Variable name(s) in parent xarray object that matches axis or coordinate `key`
     """
-
-    valid_keys = _CELL_MEASURES
-    if key not in valid_keys:
-        raise KeyError(
-            f"cf_xarray did not understand key {key!r}. Expected one of {valid_keys!r}"
-        )
 
     if isinstance(obj, DataArray):
         obj = obj._to_temp_dataset()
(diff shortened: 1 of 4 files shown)


cf_xarray/accessor.py
```diff
@@ -653,6 +647,9 @@ class CFAccessor:
 
     def __init__(self, da):
         self._obj = da
+        self._defined_cell_measures = tuple(
```
dcherian, line 650: This will take time for datasets with lots of variables every time the accessor is created.

Let's instead do

``` python
self._defined_cell_measures=None
```
and use the following when we do actually need the measure names

```python
def _get_all_cell_measures(self):
    if not self._defined_cell_measures:
        self._defined_cell_measures = set(_CELL_MEASURES + tuple(self.cell_measures))

    return self._defined_cell_measures
```

dcherian (commented):
Looks great. I have one minor suggestion. If you fix that and add a whats-new entry, this should be good to go.

cf_xarray/tests/test_accessor.py
```diff
@@ -52,20 +52,28 @@ def test_coordinates():
...
-def test_cell_measures():
+def test_cell_measures(capsys):
     ds = airds.copy(deep=True)
     ds["foo"] = xr.DataArray(ds["cell_area"], attrs=dict(standard_name="foo_std_name"))
     ds["air"].attrs["cell_measures"] += " foo_measure: foo"
     assert "foo_std_name" in ds.cf["air_temperature"].cf
 
     ds["air"].attrs["cell_measures"] += " volume: foo"
-    expected = dict(area=["cell_area"], volume=["foo"])
+    expected = dict(area=["cell_area"], foo_measure=["foo"], volume=["foo"])
```
dcherian, line 62: can we also test `ds.cf["foo_measure"]`

cf_xarray/accessor.py
malmans2: Good point. See 36de5e3400e4106696c31480b407546f58dc8c6e

cf_xarray/tests/test_accessor.py
malmans2: After #117 I think we can allow cell_measures in `ds.cf.keys`? See c45cddc35bcfbcf626ad7af72371e39160eed09b

cf_xarray/accessor.py
```diff
@@ -865,8 +874,7 @@ def keys(self) -> Set[str]:
         """
 
         varnames = list(self.axes) + list(self.coordinates)
-        if not isinstance(self._obj, Dataset):
-            varnames.extend(list(self.cell_measures))
+        varnames.extend(list(self.cell_measures))
```
dcherian, line 877: great! thanks.

dcherian (approved):
looks great. thanks @malmans2 

please go ahead and merge

PR #120, opened 2020-12-09
Remove grid_longitude from longitude standard names

aulemahal:
As noted in pangeo-data/xESMF#55 and per CF conventions, it is incorrect to assume a coordinate that has a 'grid_longitude' standard name is indeed a 'longitude' coordinate. For example, this standard name is also used by rotated pole coordinates (usually 'rlon' and 'rlat').
This PR removes the `grid_X` standard names.


dcherian:
Thanks!

PR #145, opened 2021-01-28
implement names and labels

### LINKED ISSUE #143: Drop is not working
Looks like drop is not working... Is it supposed to work or is not implemented?

```python
import cf_xarray
import xarray as xr

airds = xr.tutorial.open_dataset("air_temperature")
airds.air.attrs["standard_name"] = "air_temperature"

# Fails: Drop using standard name
airds.cf.drop("air_temperature")

# Works: Drop using variable name
airds.cf.drop("air")
```
Error is: `ValueError: One or more of the specified variables cannot be found in this dataset`

malmans2:
Closes #143 

Need to add test for:

- [x] set_coords
- [x] reset_coords
- [x] drop_vars
- [x] drop_sel
- [x] drop

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -353,6 +353,23 @@ def _get_measure(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     return list(results)
 
 
+def _get_with_key(obj: Union[DataArray, Dataset], key: str) -> List[str]:
+
+    try:
+        results = set(_get_axis_coord(obj, key))
+    except KeyError:
+        # Bypass
+        results = set()
+
+    for func in (_get_measure, _get_with_standard_name):
+        results.update(func(obj, key))
+
+    if not results:
+        raise KeyError(f"cf_xarray did not understand key {key!r}.")
+
+    return list(results)
+
+
 def _get_with_standard_name(
     obj: Union[DataArray, Dataset], name: Union[str, List[str]]
 ) -> List[str]:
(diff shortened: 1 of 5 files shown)


cf_xarray/accessor.py
```diff
@@ -376,7 +393,8 @@ def _get_with_standard_name(
     "dims_dict": (_get_axis_coord,),  # swap_dims, rename_dims
     "shifts": (_get_axis_coord,),  # shift, roll
     "pad_width": (_get_axis_coord,),  # shift, roll
-    # "names": something_with_all_valid_keys? # set_coords, reset_coords
+    "names": (_get_with_key,),  # set_coords, reset_coords, drop_vars
```
dcherian, line 396: I think just `(_get_axis_coord, _get_measure, _get_with_standard_name)` will work now.

malmans2: Yup, you're right!

malmans2:
Mmm I'm not understanding something about the CI. Why is upstream-dev push failing but pull request is OK?
It's passing on my local machine though.

cf_xarray/tests/test_accessor.py
```diff
@@ -665,3 +666,42 @@ def test_standard_name_mapper():
...
+
+    # DataArray object has no attribute set_coords
+    if not isinstance(obj, Dataset) and attr == "set_coords":
+        return
+
+    # Get attribute
+    expected = getattr(obj, attr)
+    actual = getattr(obj.cf, attr)
+
+    if attr == "drop_sel":
```
dcherian, line 685: I think it would be cleaner if we tested `drop_sel` and `reset_coords` in a different function without parmaeterizing `attr`

``` python
obj.reset_coords(...)
...
obj.drop_sel(Y=50)
```

dcherian: Let's also suppress this warning
```
dropping variables using `drop` will be deprecated; using drop_vars is encouraged.
```

malmans2: OK! I've suppressed both the Future and PendingDeprecation warnings as they are triggered by drop in the tests.

dcherian:
Thanks Mattia!

PR #147, opened 2021-01-29
add coordinate criteria table in docs

### LINKED ISSUE #138: Document coordinate criteria.
Ideally we should have a script that parses the `coordinate_criteria` dictionary and outputs a table in `.rst` format that would get rendered during the docs build. Don't know how feasible this is :)

malmans2:
Close #138: I'd also find it very useful so I gave it a go.

I've added a script to create a table with coordinate criteria in the documentation.
Not sure if I placed the table and the script in the right place (maybe the *script should just go in `doc/`?), but that should be easy to change.

Right now it creates this.

attribute_table

### DIFF
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -32,6 +32,11 @@ This dictionary contains criteria for identifying axis and coords using CF attri
 
    ~accessor.coordinate_criteria
 
+.. csv-table::
+   :file: _build/csv/coordinate_criteria.csv
+   :header-rows: 1
+   :stub-columns: 1
+
 Classes
 ~~~~~~~
 
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -33,7 +33,7 @@ This dictionary contains criteria for identifying axis and coords using CF attri
    ~accessor.coordinate_criteria
 
 .. csv-table::
-   :file: _build/csv/coordinate_criteria.csv
+   :file: _build/csv/all_criteria.csv
    :header-rows: 1
    :stub-columns: 1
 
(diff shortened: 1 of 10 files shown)


doc/contributing.rst
```diff
@@ -32,6 +32,11 @@ This dictionary contains criteria for identifying axis and coords using CF attri
 
    ~accessor.coordinate_criteria
 
+.. csv-table::
+   :file: _build/csv/coordinate_criteria.csv
```
malmans2, line 36: pre-commit is failing because this doesn't exist, it's created by `conf.py`

dcherian: I think we add `csv-table` to `ignore_directives` at the end of `setup.cfg`

dcherian:
Thanks @malmans2 this is a great improvement!

My only suggestion is to make the table more prominent. Can you put it in the top bar so it's easy to find? We can rearrange things later (maybe we want the top bar to say "Reference" and then have "API reference" and "Coordinate Criteria" as subsections)

dcherian:
It would be nice to add a link to that `introduction.ipynb` notebook

malmans2:
What about CF Criteria in the top bar, splitting axes and coordinates?

top_bar

dcherian:
That looks nice. How about transposing the table since number of rows < number of columns?

dcherian:
Should also a whats-new note for this awesome change.!

dcherian:
Could also add a table of regex patterns used for guessing

malmans2:
Regex table added!

regex

dcherian:
This is really great. Thanks!

PR #149, opened 2021-01-31
Better links in whats-new

dcherian:


setup.cfg
```diff
@@ -116,6 +116,6 @@ test = pytest
 nobeep = True
 
 [rstcheck]
-ignore_roles=pr,issue,meth
+ignore_roles=pr,issue,py:meth,py:attr
```
dcherian, line 119: this really should not be needed!

PR #158, opened 2021-02-03
Support GroupBy iteration and arithmetic

### LINKED ISSUE #157: wrapping breaks iterating over GroupBy objects
``` python
import cf_xarray

grouped = airds.cf.groupby_bins("latitude", bins=np.arange(15, 50, 10))
for label, group in grouped:
    print(label)
```

```
TypeError                                 Traceback (most recent call last)
<ipython-input-263-16beaad5fcaf> in <module>
      3 
      4 grouped = airds.cf.groupby_bins("lat", bins=np.arange(15, 50, 10))
----> 5 for label, group in grouped:
      6     print(label)

TypeError: '_CFWrappedClass' object is not iterable

```

dcherian:
Fixes #157

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -19,6 +19,7 @@ from typing import (
 
 import xarray as xr
 from xarray import DataArray, Dataset
+from xarray.core.arithmetic import SupportsArithmetic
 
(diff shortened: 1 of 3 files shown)


dcherian:
messy because we're using a private import but should be OK for now.

PR #150, opened 2021-01-31
show coords std names

malmans2:
Tiny change.
In this case:
```python
from cf_xarray.tests.datasets import airds
airds.rename_dims(lon="x", lat="y").reset_coords().cf.describe()
```
currently `longitude` and `latitude` are not shown by `.describe()`.
This change shows them in the Standard Names section.

PR #162, opened 2021-02-03
fix cell measures __repr__

malmans2:
Oops sorry, I made a mistake in #150 .
Additional cell measures were not shown in `__repr__`.

Default measures (area and volume) are now shown first, then additional measures are shown in alphabetical order.
The n/a row only applies to area and volume.

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1010,15 +1010,19 @@ class CFAccessor:
         coords = self._obj.coords
         dims = self._obj.dims
 
-        def make_text_section(subtitle, vardict, valid_values, valid_keys=None):
+        def make_text_section(subtitle, vardict, valid_values, default_keys=None):
 
             star = " * "
             tab = len(star) * " "
             subtitle = f"- {subtitle}:"
 
-            # Sort keys: Valid keys + additional in alphabetical order
-            valid_keys = [] if not valid_keys else list(valid_keys)
-            ordered_keys = valid_keys + sorted(set(vardict) - set(valid_keys))
+            # Sort keys: Order alphabetically if extra keys are available,
+            # otherwise preserve default keys order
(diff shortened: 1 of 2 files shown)


cf_xarray/tests/test_accessor.py
```diff
@@ -157,6 +157,7 @@ def test_cell_measures():
     expected = """\
     Data Variables:
     - Cell Measures:   volume: ['foo']
```
dcherian, line 159: shall we sort these alphabetically?

malmans2: Sounds good!

cf_xarray/accessor.py
```diff
@@ -1010,15 +1010,19 @@ def __repr__(self):
...
 
-            # Sort keys: Valid keys + additional in alphabetical order
-            valid_keys = [] if not valid_keys else list(valid_keys)
-            ordered_keys = valid_keys + sorted(set(vardict) - set(valid_keys))
+            # Sort keys: Order alphabetically if extra keys are available,
+            # otherwise preserve default keys order
+            default_keys = [] if not default_keys else list(default_keys)
+            extra_keys = list(set(vardict) - set(default_keys))
+            ordered_keys = (
+                sorted(default_keys + extra_keys) if extra_keys else default_keys
```
dcherian, line 1024: sorry for not being clear, how about `default_keys + sorted(extra_keys)` so there's some consistency.

malmans2: That would bring it back to the initial behavior I think. Missing keys are appended at the bottom.
Do you mean this?
```
area: n/a
volume: ['foo']
foo_measure: ['foo']
```
Would work well as long as we keep only 2 default measures, but with more `meas1, meas2, meas3: n/a` wouldn't work.

dcherian:
Ok let's merge this then

PR #164, opened 2021-02-04
Do not get bounds when extracting a DataArray

### LINKED ISSUE #144: question about time_bnds missing warning
Another question:

I am unclear about why I am getting a warning about missing time_bnds. I don't think that I need it (?)  Here is what happens:

image

image

It also happens with `cam_ts.cf.data_vars["TS"]`.

If I remove the time dimension, then I don't have a problem:
`t0 = cam_ts.isel(time=0)`
`t0.cf["TS"]`
but in many cases I need the time dimensions...

Let me know if I need to provide more information, and thanks again for the help.

malmans2:
Closes #144

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -12,6 +12,7 @@ from typing import (
     List,
     Mapping,
     MutableMapping,
+    Optional,
     Set,
     Tuple,
     Union,
@@ -609,7 +610,7 @@ def _getitem(
     try:
         for name in allnames:
             extravars = accessor.get_associated_variable_names(
-                name, get_bounds=not scalar_key
+                name, skip_bounds=scalar_key
             )
             coords.extend(itertools.chain(*extravars.values()))
 
@@ -1209,7 +1210,7 @@ class CFAccessor:
         return {k: sorted(v) for k, v in vardict.items()}
 
     def get_associated_variable_names(
-        self, name: Hashable, get_bounds=True
+        self, name: Hashable, skip_bounds: Optional[bool] = None
     ) -> Dict[str, List[str]]:
         """
         Returns a dict mapping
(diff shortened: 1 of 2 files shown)


cf_xarray/accessor.py
```diff
@@ -608,10 +608,9 @@ def check_results(names, k):
 
     try:
         for name in allnames:
-            extravars = accessor.get_associated_variable_names(name)
-            # we cannot return bounds variables with scalar keys
-            if scalar_key:
-                extravars.pop("bounds")
+            extravars = accessor.get_associated_variable_names(
```
dcherian, line 611: shall we instead suppress the warning here using the warnings module. There's probably a way to catch the warning and issue it later after checking `scalar_key`

malmans2: Yes I think that's easy to do but we should raise 2 separate warnings. One for bounds and the other for anything else. That way we still raise a warning if for example lat/lon are missing.

malmans2: https://github.com/xarray-contrib/cf-xarray/blob/48a94858111d302bdb8d72863dc58b6751283a36/cf_xarray/accessor.py#L1266-L1270

dcherian: what if we added the attribute names to warning. Then we could raise the warning `if "bounds" in warning_message and scalar_key`

malmans2: That would help but if we use one warning only we have to:
1. Catch the warning
2. Suppress it if only bounds are in the warning
3. Edit the warning text removing bounds if other variables are missing and issue the new warning

dcherian: Oh I didn't realize (3). 

Hmm.. lets use the kwarg then but rename it to `skip_bounds`?

PR #165, opened 2021-02-04
Use standard_name mapper in rename_like

### LINKED ISSUE #134: Match standard_name in .cf.rename_like
Update here:
https://github.com/xarray-contrib/cf-xarray/blob/674751b59fc5fe32759eda145b1cb6c917633367/cf_xarray/accessor.py#L1220-L1223

dcherian:
Closes #134

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2,7 +2,7 @@ import functools
 import inspect
 import itertools
 import warnings
-from collections import ChainMap
+from collections import ChainMap, Counter
 from typing import (
     Any,
     Callable,
@@ -1298,8 +1298,6 @@ class CFAccessor:
         ``other``, that variable will be renamed to match the "longitude" variable in
         ``other``.
 
-        For now, this function only matches ``("latitude", "longitude", "vertical", "time")``
-
         Parameters
         ----------
         other: DataArray, Dataset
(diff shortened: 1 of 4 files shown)


cf_xarray/tests/test_accessor.py
```diff
@@ -214,8 +214,10 @@ def test_getitem_ancillary_variables():
 def test_rename_like():
     original = popds.copy(deep=True)
 
-    with pytest.raises(KeyError):
-        popds.cf.rename_like(airds)
+    # it'll match for axis: X (lon, nlon) and coordinate="longitude" (lon, TLONG)
```
dcherian, line 217: We used to only check "coordinates", now we check all keys...

cf_xarray/accessor.py
```diff
@@ -1312,18 +1310,21 @@ def rename_like(
         ourkeys = self.keys()
         theirkeys = other.cf.keys()
 
-        good_keys = set(_COORD_NAMES) & ourkeys & theirkeys
```
dcherian, line 1315: Should we still raise an error?

cf_xarray/tests/test_accessor.py
```diff
@@ -214,10 +214,13 @@ def test_getitem_ancillary_variables():
...
+    # so delete the axis attributes
+    newair = airds.copy(deep=True)
```
dcherian, line 219: would switching to fixtures for the datasets get around this?

cf_xarray/accessor.py
```diff
@@ -1312,18 +1310,21 @@ def rename_like(
...
-            ours = _get_axis_coord_single(self._obj, key)[0]
-            theirs = _get_axis_coord_single(other, key)[0]
-            renamer[ours] = theirs
+            ours = apply_mapper(
+                (_get_axis_coord, _get_with_standard_name), self._obj, key, error=False
+            )
+            theirs = apply_mapper(
+                (_get_axis_coord, _get_with_standard_name), other, key, error=False
+            )
+            if len(ours) > 1 or len(theirs) > 1:
```
dcherian, line 1322: Could optionally raise a warning here if `verbose=True`

malmans2: Maybe warn ant the end about all `other.variables` missing in `renamer.values()`?

malmans2: This might work out with staggered grids: Rename axes first, if `len(ours)>1 or len(theirs)>1` and all variables in `ours` and `theirs` have different combinations of dimensions we should be able to rename. As we already renamed the dimensions, we just have to look at their dims.

PR #174, opened 2021-02-11
Rework mappers to be more general

### LINKED ISSUE #107: Preferred way to deal with indexed coordinates?
What is the preferred way to deal with the following situation? 

image

So, a specific time is chosen in the model output. In the first example, the time is retained in the DataArray as a coordinate that has no dimension (see next pic)

image

When the time is retained, the DataArray has a "T" axis picked up by cf-xarray. However, there isn't actually a dimension of time in this DataArray, so I would get an error if I tried to transpose the T array based on what is returned from `cf.axes`.

In the second case in the first pic, I can drop the coordinate that was indexed on with `drop=True` to avoid the "T" axis being identified. However, this is sort of annoying to need to make sure people do, and it then also loses the information of which time is being used.

Is there a reasonable around this in cf-xarray? Should an Axis or Coordinate not be identified if it has no shape/dimension/something?

malmans2:
- [x] closes #175, closes #107 
- [x] Add more tests
- [x] Copy over #165 tests
- [ ] Explicitly add conflicting functions

@dcherian were you thinking something along these lines?

Currently `_get_dims/coords/indexes` use `_get_axis_coord` and `_get_with_standard_name`.
We can probably get rid of `_get_axis_coord` and `_get_with_standard_name`, I just wanted to check first whether that's the way to go as they are used in many places. Not sure if all `_get_axis_coord` can be replaced by the same function (e.g., `_get_coords`) or it's not that simple.

All tests are passing, I didn't add new tests yet.

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -208,6 +208,19 @@ def apply_mapper(
     return results
 
 
+def _set_doc(doc):
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            return func(*args, **kwargs)
+
+        wrapper.__doc__ = doc
+        return cast(F, wrapper)
+
+    return decorator
+
+
+@_set_doc("Time variable accessor e.g. 'T.month'")
 def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List[str]:
     """
     Helper method for when our key name is of the nature "T.month" and we want to
@@ -231,18 +244,13 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
     if "." in key:
         key, ext = key.split(".", 1)
 
-        results = apply_mapper(
-            (_get_axis_coord, _get_with_standard_name), var, key, error=False
-        )
+        results = apply_mapper((_get_all,), var, key, error=False)
         if len(results) > 1:
             raise KeyError(f"Multiple results received for {key}.")
         return [v + "." + ext for v in results]
 
     else:
         return []
-
-
-_get_groupby_time_accessor.__doc__ = "Time variable accessor e.g. 'T.month'"
 
 
 def _get_axis_coord(var: Union[DataArray, Dataset], key: str) -> List[str]:
@@ -343,11 +351,6 @@ def _get_measure(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     return list(results)
 
 
-_get_measure.__doc__ = (
-    f"One or more of {_CELL_MEASURES!r};" "\n\t\t\tor arbitraty measures"
-)
-
-
 def _get_with_standard_name(
     obj: Union[DataArray, Dataset], name: Union[str, List[str]]
 ) -> List[str]:
@@ -363,37 +366,29 @@ def _get_with_standard_name(
     return varnames
 
 
+@_set_doc(
+    f"One or more of {(_AXIS_NAMES + _COORD_NAMES + _CELL_MEASURES)!r},"
+    "\n\t\t\tor arbitraty measures, or standard names"
+)
 def _get_all(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     all_mappers = (_get_axis_coord, _get_measure, _get_with_standard_name)
     results = apply_mapper(all_mappers, obj, key, error=False, default=None)
     return results
 
 
-_get_all.__doc__ = (
-    f"One or more of {(_AXIS_NAMES + _COORD_NAMES + _CELL_MEASURES)!r};"
-    "\n\t\t\tor arbitraty measures, or standard names"
-)
-
-
+@_set_doc(_get_all.__doc__ + " present in .dims")
 def _get_dims(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     return [k for k in _get_all(obj, key) if k in obj.dims]
 
 
-_get_dims.__doc__ = _get_all.__doc__ + " present in .dims"
-
-
+@_set_doc(_get_all.__doc__ + " present in .indexes")
 def _get_indexes(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     return [k for k in _get_all(obj, key) if k in obj.indexes]
 
 
-_get_indexes.__doc__ = _get_all.__doc__ + " present in .indexes"
-
-
+@_set_doc(_get_all.__doc__ + " present in .coords")
 def _get_coords(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     return [k for k in _get_all(obj, key) if k in obj.coords]
-
-
-_get_coords.__doc__ = _get_all.__doc__ + " present in .coords"
 
 
 def _variables(func: F) -> F:
@@ -835,7 +830,7 @@ class _CFWrappedPlotMethods:
             obj=self._obj.plot,
             attr=attr,
             accessor=self.accessor,
-            key_mappers=dict.fromkeys(self._keys, (_single(_get_all),)),
+            key_mappers=dict.fromkeys(self._keys, (_single(_get_coords),)),
             # TODO: "extra_decorator" is more complex than I would like it to be.
             # Not sure if there is a better way though
             extra_decorator=self._plot_decorator,
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -15,7 +15,9 @@ from typing import (
     Optional,
     Set,
     Tuple,
+    TypeVar,
     Union,
+    cast
 )
 
 import xarray as xr
@@ -144,6 +146,9 @@ ATTRS["vertical"] = ATTRS["Z"]
 
 # Type for Mapper functions
 Mapper = Callable[[Union[DataArray, Dataset], str], List[str]]
+
+# Type for decorators
+F = TypeVar("F", bound=Callable[..., Any])
 
 
 def apply_mapper(
@@ -370,38 +375,38 @@ _get_all.__doc__ = (
 )
 
 
-def _get_dims(obj: Union[DataArray, Dataset], key: str):
+def _get_dims(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     return [k for k in _get_all(obj, key) if k in obj.dims]
 
 
 _get_dims.__doc__ = _get_all.__doc__ + " present in .dims"
 
 
-def _get_indexes(obj: Union[DataArray, Dataset], key: str):
+def _get_indexes(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     return [k for k in _get_all(obj, key) if k in obj.indexes]
 
 
 _get_indexes.__doc__ = _get_all.__doc__ + " present in .indexes"
 
 
-def _get_coords(obj: Union[DataArray, Dataset], key: str):
+def _get_coords(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     return [k for k in _get_all(obj, key) if k in obj.coords]
 
 
 _get_coords.__doc__ = _get_all.__doc__ + " present in .coords"
 
 
-def _variables(func):
+def _variables(func: F) -> F:
     @functools.wraps(func)
-    def get_variables(obj: Union[DataArray, Dataset], key: str):
+    def wrapper(obj: Union[DataArray, Dataset], key: str) -> List[DataArray]:
         return [obj[k] for k in func(obj, key)]
 
-    return get_variables
+    return cast(F, wrapper)
 
 
-def _single(func):
+def _single(func: F) -> F:
     @functools.wraps(func)
-    def get_single(obj: Union[DataArray, Dataset], key: str):
+    def wrapper(obj: Union[DataArray, Dataset], key: str):
         results = func(obj, key)
         if len(results) > 1:
             raise KeyError(
@@ -411,9 +416,9 @@ def _single(func):
             raise KeyError(f"No results found for {key!r}.")
         return results
 
-    get_single.__doc__ = func.__doc__.replace("One or more of", "One of")
+    wrapper.__doc__ = func.__doc__.replace("One or more of", "One of") if func.__doc__ else func.__doc__
 
-    return get_single
+    return cast(F, wrapper)
 
 
 #: Default mappers for common keys.
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -436,7 +436,7 @@ _DEFAULT_KEY_MAPPERS: Mapping[str, Tuple[Mapper, ...]] = {
     "labels": (_get_indexes,),  # drop_sel
     "coords": (_get_dims,),  # interp
     "indexers": (_get_dims,),  # sel, isel, reindex
-    # "indexes": (_get_axis_coord,),  # set_index
+    "indexes": (_get_dims,),  # set_index TODO: cf_xarray decodes keys, not values
     "dims_or_levels": (_get_dims,),  # reset_index
     "window": (_get_dims,),  # rolling_exp
     "coord": (_single(_get_coords),),  # differentiate, integrate
@@ -617,12 +617,12 @@ def _getitem(
     successful = dict.fromkeys(key, False)
     for k in key:
         if "coords" not in skip and k in _AXIS_NAMES + _COORD_NAMES:
-            names = _get_axis_coord(obj, k)
+            names = _get_all(obj, k)
             check_results(names, k)
             successful[k] = bool(names)
             coords.extend(names)
         elif "measures" not in skip and k in accessor._get_all_cell_measures():
-            measure = _get_measure(obj, k)
+            measure = _get_all(obj, k)
             check_results(measure, k)
             successful[k] = bool(measure)
             if measure:
@@ -1145,10 +1145,7 @@ class CFAccessor:
         Dictionary of valid Axis names that can be used with ``__getitem__`` or ``.cf[key]``.
         Will be ("X", "Y", "Z", "T") or a subset thereof.
         """
-        vardict = {
-            key: apply_mapper(_get_axis_coord, self._obj, key, error=False)
-            for key in _AXIS_NAMES
-        }
+        vardict = {key: _get_coords(self._obj, key) for key in _AXIS_NAMES}
 
         return {k: sorted(v) for k, v in vardict.items() if v}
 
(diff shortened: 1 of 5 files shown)


dcherian:
yes! something very similar see #175. 

I would use `apply_mapper` though to apply them. and I think `_get_single` should be a decorator. So in the dictionary we would have something like `"indexes": (_single(_get_indexes))`

Can you unify the two PRs?

cf_xarray/accessor.py
```diff
@@ -430,13 +456,21 @@ def _build_docstring(func):
...
+        f"One or more of {(_AXIS_NAMES + _COORD_NAMES)!r};\n\t\t\tor standard names"
+    )
     # this list will need to be updated any time a new mapper is added
     mapper_docstrings = {
         _get_axis_coord: f"One or more of {(_AXIS_NAMES + _COORD_NAMES)!r}",
         _get_axis_coord_single: f"One of {(_AXIS_NAMES + _COORD_NAMES)!r}",
         _get_groupby_time_accessor: "Time variable accessor e.g. 'T.month'",
         _get_with_standard_name: "Standard names",
         _get_measure_variable: f"One of {_CELL_MEASURES!r}",
+        _single(_get_all): get_all_docstring.replace("One or more of", "One of"),
```
malmans2, line 469: With the decorator this doesn't work well... Looks like anytime I use a decorator the function gets a different id, so the key `_single(_get_all)` is not in ``_DEFAULT_KEY_MAPPERS``. This is the only workaround I can think of: use `func.__name__` as keys of this dictionary, and force the decorator to add an attribute/annotation so we know that `_single` was applied. When we loop later using ``_DEFAULT_KEY_MAPPERS``, we replace "One or more of" with "One of" if there is the decorator attribute/annotation. There must be a cleaner way...

dcherian: we could add these docstrings to the actual functions. and `_single` could extract `__doc__`...

This was me trying to be fancy so we could also just delete it. Up to you...

malmans2: I think `__doc__` works well

cf_xarray/accessor.py
```diff
@@ -403,31 +429,33 @@ def get_single(obj: Union[DataArray, Dataset], key: str):
...
 #: Default mappers for common keys.
 _DEFAULT_KEY_MAPPERS: Mapping[str, Tuple[Mapper, ...]] = {
-    "dim": (_get_dims,),
-    "dims": (_get_dims,),  # transpose
-    "drop_dims": (_get_dims,),  # drop_dims
-    "dimensions": (_get_dims,),  # stack
-    "dims_dict": (_get_dims,),  # swap_dims, rename_dims
-    "shifts": (_get_dims,),  # shift, roll
-    "pad_width": (_get_dims,),  # shift, roll
+    "dim": (_dims(_get_all),),
```
malmans2, line 439: How about using more decorators? Too messy?

cf_xarray/accessor.py
```diff
@@ -403,31 +429,33 @@ def get_single(obj: Union[DataArray, Dataset], key: str):
...
+    "dims_or_levels": (_dims(_get_all),),  # reset_index
+    "window": (_dims(_get_all),),  # rolling_exp
+    "coord": (_single(_coords(_get_all)),),  # differentiate, integrate
     "group": (_single(_get_all), _get_groupby_time_accessor),  # groupby
-    "indexer": (_single(_get_indexes),),  # resample
+    "indexer": (_single(_indexes(_get_all)),),  # resample
     "variables": (_get_all,),  # sortby
-    "weights": (_get_measure_variable,),  # type: ignore
-    "chunks": (_get_dims,),  # chunk
+    "weights": (_variables(_single(_get_all)),),  # type: ignore
```
malmans2, line 457: I thing we can look at all variables here, right? Before was the equivalent of `(_variables(_single(_get_measure)),),`

cf_xarray/accessor.py
dcherian: i like your odler version with `_get_dims` etc. that's nice and clean. And `_single` can be the only decorator since it is not used very frequently. `_variables` could also be a decorator I guess.

cf_xarray/accessor.py
dcherian: yes!

cf_xarray/accessor.py
malmans2: Yup, I was very undecided about it!

malmans2:
Should we try to use the new functions as much as possible? 
For example, let's use `_single(_get_all)(var, key)` rather than `_get_axis_coord_single(var, key)`?

dcherian:
Yes I think that is right. You'll basically implement #165 too :) if you change all of it. Copy the tests over?

malmans2:
Sounds good! I started with `_get_axis_coord_single` always replaced by `_single(_get_all)` (all current tests are passing, but can you check whether `_get_all` is always the right one. I think so, but I'm not 100% sure).

malmans2:
Actually, the readthedocs action caught it. Should be `_get_coords` for plots.

cf_xarray/accessor.py
```diff
@@ -835,7 +830,7 @@ def __getattr__(self, attr):
             obj=self._obj.plot,
             attr=attr,
             accessor=self.accessor,
-            key_mappers=dict.fromkeys(self._keys, (_single(_get_all),)),
+            key_mappers=dict.fromkeys(self._keys, (_single(_get_coords),)),
```
malmans2, line 833: What about `scatter`? Should we add an if statement?
Also, just `get_coords`, right? For `cf_xarray` dims and indexes should be coords as well?

cf_xarray/accessor.py
```diff
@@ -203,18 +208,19 @@ def _apply_single_mapper(mapper):
...
-    """ Helper method for when we really want only one result per key. """
-    results = _get_axis_coord(var, key)
-    if len(results) > 1:
-        raise KeyError(
-            f"Multiple results for {key!r} found: {results!r}. I expected only one."
-        )
-    elif len(results) == 0:
-        raise KeyError(f"No results found for {key!r}.")
-    return results
+def _set_doc(doc):
```
dcherian, line 211: hmm.. i think this is a little too complicated. Can we just add the docstrings for the functions directly instead? I think the redundancy is OK, I don't expect it to change...

dcherian (approved):

cf_xarray/accessor.py
```diff
@@ -372,36 +366,82 @@ def _get_with_standard_name(
     return varnames
 
 
+@_set_doc(
+    f"One or more of {(_AXIS_NAMES + _COORD_NAMES + _CELL_MEASURES)!r},"
```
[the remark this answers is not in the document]
malmans2: OK! Let's use tests to remind us to update docs, just in case we will change constants?

cf_xarray/accessor.py
malmans2: Yes, looks like `_get_all` is OK.

malmans2:
I'm not following what's happening here:
https://github.com/xarray-contrib/cf-xarray/blob/425c4720d83c97a17f509c3ce4977bcec301ef64/cf_xarray/accessor.py#L813-L837
I must use `_get_coords` in `__call__`, otherwise this fails:
```python
airds.air.cf.weighted("area").mean(["latitude", "time"]).cf.plot(x="longitude")
```
Error is ValueError: Cannot specify both x and y kwargs for line plots

However, `_get_all` is fine in `__getattr__`, and this is OK:
```
airds.air.cf.weighted("area").mean(["latitude", "time"]).cf.plot.line(x="longitude")
```

dcherian:
Phew after some debugging... 

```
>>> da = airds.air.cf.weighted("area").mean(["latitude", "time"])
>>> cf_xarray.accessor._get_with_standard_name(da, None)
['<this-array>']
```

Also
```
cf_xarray.accessor._get_with_standard_name(airds, None)
["cell_area"]
```

We were also missing one line plot case (`.plot` with 1D arrays)

cf_xarray/accessor.py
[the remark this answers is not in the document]
dcherian: good idea!

cf_xarray/accessor.py
```diff
@@ -1197,10 +1191,10 @@ def cell_measures(self) -> Dict[str, List[str]]:
                 da.attrs.get("cell_measures", "") for da in obj.data_vars.values()
             ]
 
-        measures: Dict[str, List[str]] = {}
+        keys = {}
         for attr in all_attrs:
-            for key, value in parse_cell_methods_attr(attr).items():
-                measures[key] = measures.setdefault(key, []) + [value]
+            keys.update(parse_cell_methods_attr(attr))
+        measures = {key: _get_all(self._obj, key) for key in keys}
```
malmans2, line 1197: Doesn't change anything, just seems better to use `_get_all`.

cf_xarray/accessor.py
```diff
@@ -436,7 +436,7 @@ def wrapper(obj: Union[DataArray, Dataset], key: str):
     "labels": (_get_indexes,),  # drop_sel
     "coords": (_get_dims,),  # interp
     "indexers": (_get_dims,),  # sel, isel, reindex
-    # "indexes": (_get_axis_coord,),  # set_index
+    "indexes": (_get_dims,),  # set_index TODO: cf_xarray decodes keys, not values
```
malmans2, line 439: `.set_index(longitude="lon")` would work, `.set_index(longitude="longitude")` does not work.
I should wrap it in `_single`

cf_xarray/accessor.py
```diff
@@ -617,12 +617,12 @@ def check_results(names, k):
     successful = dict.fromkeys(key, False)
     for k in key:
         if "coords" not in skip and k in _AXIS_NAMES + _COORD_NAMES:
-            names = _get_axis_coord(obj, k)
+            names = _get_all(obj, k)
```
malmans2, line 620: Maybe should be `_get_coords`, so it is consistent with `cf.coordinates`?

cf_xarray/accessor.py
```diff
@@ -617,12 +617,12 @@ def check_results(names, k):
     successful = dict.fromkeys(key, False)
     for k in key:
         if "coords" not in skip and k in _AXIS_NAMES + _COORD_NAMES:
-            names = _get_axis_coord(obj, k)
```
dcherian, line 586: we should think about this one. when pulling out dataarrays we want the appropriate 'latitude' and 'longitude'...

cf_xarray/accessor.py
```diff
@@ -1167,10 +1164,7 @@ def coordinates(self) -> Dict[str, List[str]]:
         Dictionary of valid Coordinate names that can be used with ``__getitem__`` or ``.cf[key]``.
         Will be ("longitude", "latitude", "vertical", "time") or a subset thereof.
         """
-        vardict = {
-            key: apply_mapper(_get_axis_coord, self._obj, key, error=False)
-            for key in _COORD_NAMES
-        }
+        vardict = {key: _get_coords(self._obj, key) for key in _COORD_NAMES}
```
dcherian, line 1167: We should update the docstring to say that it only searches in `.coords` similarly for .axes etc.

cf_xarray/accessor.py
dcherian: lets save this for later. this might actually need a reimplementation. We want to change the key with `_get_dims` and the value with `_get_all`?

cf_xarray/accessor.py
malmans2: What's the downside of treating coordinates as measures, so `_get_all` rather than `_get_coords`?

cf_xarray/accessor.py
malmans2: Yes exactly, not sure whether that's something easy to do. I'll add `_single` but will comment everything out.

cf_xarray/accessor.py
dcherian: Is there some weird thing where there's a "data variable" with "latitude" values but also a different latitude coordinate? Like a trajectory thing with "true value longitude" and "nominal longitude"? Should we distinguish between the two?

malmans2: Something like this?
 ```python
 _get_coords(obj, key) if _get_coords(obj, key) else _get_all(obj, key)
```

Comment (malmans2) [reply cf_xarray/accessor.py:1167]: mmm, actually the way it is right now is probably the best

Comment (dcherian) [reply cf_xarray/accessor.py:1167]: The first comment in #107 brings up this point.

The fundamental issue is whether `.axes` et al should returns results present only in coordinates or in all variables. With the methods, we can control this by using `_get_dims` and friends. With the properties we don't really have a good way of adding any user control. That  said, asking people to use `set(obj.cf.axes) & set(obj.cf.dims)` is really not too bad.

Maybe all properties should return names present in all variables (not just data variables), and we can add an FAQ showing people how to use set operations to pull out what's needed?

File: cf_xarray/accessor.py (hunk 10 above)
Comment (malmans2) [reply cf_xarray/accessor.py:586]: This is what I tried to break this (coordinate actually named latitude):
```python
from cf_xarray.datasets import popds
popds["ULAT"].attrs["standard_name"] = "latitude"
popds = popds.assign_coords(latitude=popds["ULAT"])
popds = popds.reset_coords("ULAT")
assert "latitude" not in popds.cf["UVEL"].coords
assert "ULAT" in popds.cf["UVEL"].coords
popds["UVEL"].attrs["coordinates"] = popds["UVEL"].attrs["coordinates"].replace("ULAT", "latitude")
assert "latitude" in popds.cf["UVEL"].coords
assert "ULAT" not in popds.cf["UVEL"].coords
```
Looks OK, and I can't think of other cases where this change wouldn't work. Any ideas?

Accidentally I found out that this raises an error:
```python
popds.cf.data_vars
```
Error is: Receive multiple variables for key 'latitude': {'ULAT', 'latitude'}. Expected only one. Please pass a list ['latitude'] instead to get all variables matching 'latitude'.

Not sure whether this is correct, as there's just one latitude `data_var` (but multiple `coords`), but it looks like we used to have the same behavior?

```
Coordinates:
- CF Axes: * X: ['nlon']
           * Y: ['nlat']
             Z, T: n/a

- CF Coordinates:   longitude: ['TLONG', 'ULONG']
                    latitude: ['TLAT', 'latitude']
                    vertical, time: n/a

- Cell Measures:   area, volume: n/a

- Standard Names:   latitude: ['latitude']

Data Variables:
- Cell Measures:   area, volume: n/a

- Standard Names:   latitude: ['ULAT']
                    sea_water_potential_temperature: ['TEMP']
                    sea_water_x_velocity: ['UVEL']
```

dcherian: > popds.cf.data_vars

Nice find! Definitely a bug. Let's open an issue for that.

The `ULAT` vs `latitude` issue is interesting. Fundamentally we need to decide whether cf-xarray treats xarray objects as "bags of variables" ignoring distinctions like data_vars and coords, or whether it fully follows xarray's data_vars vs coords. I'm leaning towards the former i.e. cf-xarray tries its best to attach variables mentioned in attributes ignoring any data_vars/coords distinction. I think we do this will ancillary variables for e.g. (?)


``` python
from cf_xarray.datasets import popds
popds["ULAT"].attrs["standard_name"] = "latitude"
popds = popds.assign_coords(latitude=popds["ULAT"])
popds = popds.reset_coords("ULAT")
assert "latitude" not in popds.cf["UVEL"].coords
assert "ULAT" in popds.cf["UVEL"].coords  
```

malmans2: Yes, I also think this is the best way to handle it (and I'm in favor of changing what's returned by properties such as `cf.coordinates` https://github.com/xarray-contrib/cf-xarray/issues/107#issuecomment-782908902)

malmans2:
Should we merge `_get_axis_coord` and `_get_axis_measure` so we only loop once? We still use `_get_with_standard_name` outside of `_get_all`: https://github.com/xarray-contrib/cf-xarray/blob/a3178b501fb0617f02809fd358773b2889787612/cf_xarray/accessor.py#L631

dcherian:
> Should we merge _get_axis_coord and _get_measure so we only loop once?

I'd like to keep them separate for now since they do different things. One optimization is to add `keys_to_search` as a kwarg to the mapper functions. for example,`_get_indexes` would be `_get_all(..., keys_to_search=obj.indexes)`. Then `_get_all` would only loop over `keys_to_search` instead of `obj.variables`

We can do this later though. This is a nice change already.

malmans2:
Mmmm, I don't think we are getting the right error here:
```python
from cf_xarray.datasets import popds
popds.cf.differentiate("longitude")
```
ValueError: Coordinate longitude does not exist.

The error should be raised before by `@_single`

malmans2:
Should we somehow always set this True when using `@_single`?
https://github.com/xarray-contrib/cf-xarray/blob/bc2b14d08a860e30efb3db12891d25f5231a2747/cf_xarray/accessor.py#L176-L177

dcherian:
Yes looks right

malmans2:
Actually, I think we can still allow the case with no results. Let's use a very specific message here, so we know this one should always be raised, or an error different than KeyError?
https://github.com/xarray-contrib/cf-xarray/blob/bc2b14d08a860e30efb3db12891d25f5231a2747/cf_xarray/accessor.py#L409-L412

dcherian:
KeyError seems right, what do you want to change? It looks like `longitude` didn't get mapped to anything. isn't that the issue?

malmans2:
```python
from cf_xarray.datasets import popds
popds.cf.differentiate("longitude")
```
I think longitude is mapped to 2 variables, but `_rewrite_values` switches off the error from `single`? 
https://github.com/xarray-contrib/cf-xarray/blob/bc2b14d08a860e30efb3db12891d25f5231a2747/cf_xarray/accessor.py#L975-L978
Error should be "Multiple results for 'longitude' found: ['ULONG', 'TLONG']. I expected only one."

dcherian:
I say we raise ValueError in `_single` and open an issue to come up with a better solution

malmans2:
Sounds good

dcherian:
in `apply_mapper` we could also search for "expected only one" in the exception message and raise in that case even when `error is False`

malmans2:
Yes that's my favorite solution but I think we need a slightly more specific error because "expected only one" is also used by `check_results` (although e is upper case in the latter).

malmans2:
Should be almost good to go. Unless you think we should add more tests to make sure we cover all signatures and their xarray functions, we can probably merge and explicitly add functions to the accessor later if we find conflicts.

dcherian:
LGTM. Thanks @malmans2

Issue #179, opened 2021-02-24
Accessing with standard name broken on DataArrays.


aulemahal:
So I know this is a strange call, but it did work before and doesn't anymore with version 0.5.0. I think this has to do with the refactor related to  `_get_all` and friends.

## What I did
```python
import cf_xarray
import xarray as xr

ds = xr.tutorial.open_dataset('air_temperature')
ds.lat.cf['latitude']
```

I expected to receive `ds.lat` back, as was the behaviour before. But now, an error is raised:
```python3
<ipython-input-26-1c2f69690a8b> in <module>
----> 1 ds.lat.cf['latitude']

~/Python/cf-xarray/cf_xarray/accessor.py in __getitem__(self, key)
   1723             )
   1724 
-> 1725         return _getitem(self, key)
   1726 
   1727     pass

~/Python/cf-xarray/cf_xarray/accessor.py in _getitem(accessor, key, skip)
    619         if "coords" not in skip and k in _AXIS_NAMES + _COORD_NAMES:
    620             names = _get_all(obj, k)
--> 621             check_results(names, k)
    622             successful[k] = bool(names)
    623             coords.extend(names)

~/Python/cf-xarray/cf_xarray/accessor.py in check_results(names, k)
    607     def check_results(names, k):
    608         if scalar_key and len(names) > 1:
--> 609             raise KeyError(
    610                 f"Receive multiple variables for key {k!r}: {names}. "
    611                 f"Expected only one. Please pass a list [{k!r}] "

KeyError: "Receive multiple variables for key 'latitude': ['<this-array>', 'lat']. Expected only one. Please pass a list ['latitude'] instead to get all variables matching 'latitude'."
```

The error seems to be related to `get_with_standard_name` which will return `<this-array>` in our case.  

EDIT: Then example below fails because of the bug I highlighted, but it realized it should fail nonetheless.

<details>

Another failing example:
```python3
ds.air.attrs['standard_name'] = 'stdname'

ds.cf['stdname']  #-> works and returns ds.air
ds.air.cf['stdname']
```
The last line fails with
```
KeyError: "DataArray.cf does not understand the key 'stdname'. Use 'repr(DataArray.cf)' (or 'DataArray.cf' in a Jupyter environment) to see a list of key names that can be interpreted."
```

I think `_get_with_standard_name` must be modified to not return '<this-array>', but the var name instead.

</details>

malmans2:
[...]

PR #180, opened 2021-02-24
Do not return <this-array> instances

aulemahal:
Small fix for #179.

Modified `_get_with_standard_name` so that when used on a `DataArray`, it cannot return itself (the `<this-array>` variable in the temp dataset). This match the previous behavior (I think?) and fits the returned error message : `ds.air.cf['air_temperature']` fails because 'air_temperature' doesn't appear in `ds.air.cf`.  But `ds.lat.cf['latitude']` works, since 'latitude' appears in `ds.lat.cf`.

Also added a minimal test.

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -351,10 +351,10 @@ def _get_with_standard_name(
 
     varnames = []
     if isinstance(obj, DataArray):
-        obj = obj._to_temp_dataset()
+        obj = obj.coords.to_dataset()
     for vname, var in obj.variables.items():
         stdname = var.attrs.get("standard_name", None)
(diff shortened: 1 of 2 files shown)


cf_xarray/accessor.py
```diff
@@ -354,7 +354,7 @@ def _get_with_standard_name(
         obj = obj._to_temp_dataset()
```
dcherian, line 354:
```suggestion
        obj = obj.coords.to_dataset()
```

How about this?

aulemahal: Oh, well yeah! Same result, but more elegant!

dcherian:
Great! merging. Can you test your downstream packages to make sure there are not other issues.

PR #181, opened 2021-02-26
cf.data_vars skips coords

### LINKED ISSUE #177: cf.data_vars fails when there are coords with conflicting standard names
```python
from cf_xarray.datasets import popds
popds["ULAT"].attrs["standard_name"] = "latitude"
popds["TLAT"].attrs["standard_name"] = "latitude"
popds = popds.reset_coords("ULAT")
popds.cf.data_vars
```
Error:
```
KeyError: "Receive multiple variables for key 'latitude': {'ULAT', 'TLAT'}. Expected only one. Please pass a list ['latitude'] instead to get all variables matching 'latitude'."
```

This shouldn't happen as `ULAT` is the only latitude in `data_vars`.

malmans2:
Close #177

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -639,7 +639,6 @@ def _getitem(
                 check_results(stdvars, k)
                 successful[k] = bool(stdvars)
             varnames.extend(stdvars)
-            
 
     # these are not special names but could be variable names in underlying object
     # we allow this so that we can return variables with appropriate CF auxiliary variables
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -629,11 +629,17 @@ def _getitem(
                 varnames.extend(measure)
         else:
             stdnames = set(_get_with_standard_name(obj, k))
-            check_results(stdnames, k)
-            successful[k] = bool(stdnames)
             objcoords = set(obj.coords)
-            varnames.extend(stdnames - objcoords)
-            coords.extend(stdnames & objcoords)
+            stdvars = stdnames - objcoords
+            if "coords" not in skip:
+                check_results(stdnames, k)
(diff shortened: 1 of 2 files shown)


cf_xarray/accessor.py
```diff
@@ -629,11 +629,16 @@ def check_results(names, k):
                 varnames.extend(measure)
         else:
             stdnames = set(_get_with_standard_name(obj, k))
-            check_results(stdnames, k)
-            successful[k] = bool(stdnames)
             objcoords = set(obj.coords)
-            varnames.extend(stdnames - objcoords)
-            coords.extend(stdnames & objcoords)
+            stdvars = stdnames - objcoords
```
dcherian, line 633: move stdvars into the else block and add a comment?

cf_xarray/accessor.py
```diff
@@ -630,15 +630,12 @@ def check_results(names, k):
...
-            if "coords" not in skip:
-                check_results(stdnames, k)
-                successful[k] = bool(stdnames)
-                coords.extend(stdnames & objcoords)
-            else:
-                check_results(stdvars, k)
-                successful[k] = bool(stdvars)
-            varnames.extend(stdvars)
+            if "coords" in skip:
+                stdnames -= objcoords
```
malmans2, line 634: How about this, exactly the same as the origin with an extra if statement?

malmans2:
I don't think the error in the doc notebook is related to this PR. Opening a separate issue.

cf_xarray/accessor.py
dcherian: Perfect.

dcherian:
Looks like the demo notebook is now broken.

dcherian:
ah just saw your comment.

PR #183, opened 2021-02-26
rename_like warns about conflicting variables

### LINKED ISSUE #182: Documentation notebook is broken
Looks like there are two issues:

1. For some reason, the latest rtd build does not show any output after the first error (cell 11, the exception tag looks OK)
2. Cell 23 is broken: `pop.cf["TEMP"].cf.rename_like(ds)`, and I think we didn't catch it in previous PRs because of point 1.

`pop.cf["TEMP"].cf.rename_like(ds)` fails because we are trying to rename:
```
Coordinates:
- CF Axes: * X: ['nlon']
           * Y: ['nlat']
             Z, T: n/a

- CF Coordinates:   longitude: ['TLONG']
                    latitude: ['TLAT']
                    vertical, time: n/a

- Cell Measures:   area, volume: n/a

- Standard Names:   n/a
```
to
```
Coordinates:
- CF Axes: * X: ['lon']
           * Y: ['lat']
           * T: ['time']
             Z: n/a

- CF Coordinates: * longitude: ['lon']
                  * latitude: ['lat']
                  * time: ['time']
                    vertical: n/a

- Cell Measures:   area, volume: n/a

- Standard Names: * latitude: ['lat']
                  * longitude: ['lon']
                  * time: ['time']

Data Variables:
- Cell Measures:   area, volume: n/a

- Standard Names:   air_temperature: ['air']
```
We get an error because we need to rename both `nlon` and `TLONG` to `lon`.
Maybe @dcherian already took care of this in #165?

malmans2:
Closes #182

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1348,13 +1348,12 @@ class CFAccessor:
 
         conflicts = {}
         for k0, v0 in keydict.items():
+            if len(v0["ours"]) > 1 or len(v0["ours"]) > 1:
+                conflicts[k0] = v0
+                continue
             for v1 in keydict.values():
                 # Conflicts have same ours but different theirs or vice versa
-                if (
-                    sum([v0["ours"] == v1["ours"], v0["theirs"] == v1["theirs"]]) == 1
-                    or len(ours) > 1
-                    or len(theirs) > 1
-                ):
+                if sum([v0["ours"] == v1["ours"], v0["theirs"] == v1["theirs"]]) == 1:
                     conflicts[k0] = v0
                     break
         if conflicts:
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1347,15 +1347,13 @@ class CFAccessor:
             renamer[ours] = theirs
 
         conflicts = {
-            k: renamer[k]
-            for k in sorted(renamer)
-            if list(renamer.values()).count(renamer[k]) > 1
+            k: v for k, v in renamer.items() if list(renamer.values()).count(v) > 1
         }
         if conflicts:
             renamer = {k: v for k, v in renamer.items() if k not in conflicts}
             warnings.warn(
                 "Conflicting variables skipped:\n"
-                + "\n".join([f"{k}: {v}" for k, v in conflicts.items()]),
+                + "\n".join([f"{k}: {conflicts[k]}" for k in sorted(conflicts)]),
                 UserWarning,
             )
 
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1348,7 +1348,7 @@ class CFAccessor:
 
         conflicts = {}
         for k0, v0 in keydict.items():
-            if len(v0["ours"]) > 1 or len(v0["ours"]) > 1:
+            if len(v0["ours"]) > 1 or len(v0["theirs"]) > 1:
                 conflicts[k0] = v0
                 continue
             for v1 in keydict.values():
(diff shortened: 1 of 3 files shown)


cf_xarray/accessor.py
```diff
@@ -1346,6 +1346,19 @@ def rename_like(
             theirs = _single(_get_all)(other, key)[0]
             renamer[ours] = theirs
```
malmans2, line 1347: Are there weird cases where this is actually a conflict and we should skip and warn about it?

dcherian: no idea ;) lets find out.

cf_xarray/accessor.py
```diff
@@ -1346,6 +1346,17 @@ def rename_like(
...
 
+        conflicts = {
+            k: v for k, v in renamer.items() if list(renamer.values()).count(v) > 1
+        }
+        if conflicts:
+            renamer = {k: v for k, v in renamer.items() if k not in conflicts}
+            warnings.warn(
+                "Conflicting variables skipped:\n"
+                + "\n".join([f"{k}: {conflicts[k]}" for k in sorted(conflicts)]),
```
dcherian, line 1356: From the doc notebook example
```
da = pop.cf["TEMP"]
da.cf.rename_like(ds)
```
the warning says
```
TLAT: lat
TLONG: lon
nlat: lat
nlon: lon
```

I'd like this to say

```
TLAT: lat ("latitude")
TLONG: lon  ("longitude")
nlat: lat ("Y")
nlon: lon ("X")
```

dcherian (commented):
Thanks Mattia.

It'd be nice if there was a way for the user to limit the renaming in some form but maybe this should be a future PR

```
pop.cf["TEMP"].cf.rename_like(ds, do=["coordinates", "standard_name"])
```

This would work, since it would skip `nlon`, `nlat` so there would be no conflicts.

Likewise
```
pop.cf["TEMP"].cf.rename_like(ds, do=["axes"])
```

woudl rename `nlon`, `nlat` to `lon`, `lat`

cf_xarray/accessor.py
```diff
@@ -1340,23 +1340,36 @@ def rename_like(
...
-        conflicts = {
-            k: v for k, v in renamer.items() if list(renamer.values()).count(v) > 1
-        }
+            keydict[key] = dict(ours=ours, theirs=theirs)
+
+        conflicts = {}
+        for k0, v0 in keydict.items():
+            for v1 in keydict.values():
+                # Conflicts have same ours but different theirs or vice versa
+                if sum([v0["ours"] == v1["ours"], v0["theirs"] == v1["theirs"]]) == 1:
```
malmans2, line 1353: This is actually also addressing this: https://github.com/xarray-contrib/cf-xarray/pull/183/files/2a7d312dfa877a624ecdd837029559901fde9cc7#r583839911

cf_xarray/accessor.py
```diff
@@ -1340,12 +1340,36 @@ def rename_like(
         theirkeys = other.cf.keys()
 
         good_keys = ourkeys & theirkeys
-        renamer = {}
+        keydict = {}
         for key in good_keys:
             ours = _single(_get_all)(self._obj, key)[0]
```
malmans2, line 1345: Should we catch the error and just skip+warn?

dcherian: yes that sounds right

malmans2:
Looks like this now:
```python
from cf_xarray.datasets import airds, popds
popds.cf.rename_like(airds)
```
```
UserWarning: Conflicting variables skipped:
['TLAT', 'ULAT']: ['lat'] (latitude)
['TLONG', 'ULONG']: ['lon'] (longitude)
['nlat']: ['lat'] (Y)
['nlon']: ['lon'] (X)
```

dcherian:
Thanks @malmans2

PR #196, opened 2021-04-05
Updates to getitem

### LINKED ISSUE #191: lat/lon_bnds with units attribute identified as latitude/longitude coordinate variables when registered as coords of the Dataset
Hello,

I encountered a minor problem when trying to identify the latitude/longitude coordinate variables of an xarray.Dataset. I used version 0.5.1 of cf_xarray.

#### Description
When the latitude bounds and longitude bounds are registered as coordinates of the xarray.Dataset and they have a units attribute ("degrees_north/east") cf_xarray identifies them as latitude / longitude coordinate variables, which I did not expect.

The latitude bounds and longitude bounds get registered as coordinates of the xarray.Dataset for example when calling `xarray.open_dataset(path_to_ds, decode_coords="all")` .

I'm not sure, but I think CF conventions allow for lat_bnds/lon_bnds to have a units attribute. At least I know that such CMIP6 datasets passed a quality control including a CF-checker.

#### Short example

```python
import xarray as xr
import cf_xarray as cfxr
import numpy as np

lons=np.zeros((10))
lats=np.zeros((10))
lat_bounds=np.zeros((10,2))
lon_bounds=np.zeros((10,2))

ds = xr.Dataset(data_vars={},
                coords={'lat': ('lat', lats),
                        'lon': ('lon', lons),                      
                        'lat_bnds': (['lat', 'bnds'], lat_bounds),
                        'lon_bnds': (['lon', 'bnds'], lon_bounds)})


ds["lat"].attrs = {
    "bounds": "lat_bnds",
    "units": "degrees_north",
    "long_name": "latitude",
    "standard_name": "latitude",
    "axis": "Y",
}
ds["lon"].attrs = {
    "bounds": "lon_bnds",
    "units": "degrees_east",
    "long_name": "longitude",
    "standard_name": "longitude",
    "axis": "X",
}
ds["lat_bnds"].attrs = {
    "long_name": "latitude_bounds",
    "units": "degrees_north",
}
ds["lon_bnds"].attrs = {
    "long_name": "longitude_bounds",
    "units": "degrees_east",
}

print(ds.cf["latitude"].name)
```

That results in the following error as it also identifies lat_bnds for the key "latitude".

```python
KeyError: "Receive multiple variables for key 'latitude': ['lat_bnds', 'lat']. Expected only one. Please pass a list ['latitude'] instead to get all variables matching 'latitude'."
```

Is it the desired behaviour to list all coordinate variables associated with the latitude coordinate (incl. bounds), and not only the latitude coordinate variable itself?

As a workaround for my problem I can use for example `cf_xarray.accessor._get_with_standard_name` or `xarray.reset_coords()`.

### LINKED ISSUE #195: bad "cell_measures" attribute breaks things.
I think we should raise a warning in this case.

```
ValueError                                Traceback (most recent call last)
<ipython-input-23-18478435d3d6> in <module>
----> 1 cmor.cf["sea_water_x_velocity"]
...

~/python/cf-xarray/cf_xarray/utils.py in parse_cell_methods_attr(attr)
     39     strings = [s for scolons in attr.split(":") for s in scolons.split()]
     40     if len(strings) % 2 != 0:
---> 41         raise ValueError(f"attrs['cell_measures'] = {attr!r} is malformed.")
     42 
     43     return dict(zip(strings[slice(0, None, 2)], strings[slice(1, None, 2)]))

ValueError: attrs['cell_measures'] = '--OPT' is malformed.
```

dcherian:
- ignore bad cell_measures
- ignore bounds variable when provided a scalar key
 
Closes #195
Closes #191

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -605,12 +605,22 @@ def _getitem(
     if skip is None:
         skip = []
 
-    def check_results(names, k):
+    def drop_bounds(names):
+        # sometimes bounds variables have the same standard_name as the
+        # actual variable. It seems practical to ignore them when indexing
+        # with a scalar key. Hopefully these will soon get decoded to IntervalIndex
+        # and we can move on...
+        if scalar_key:
+            bounds = set([obj[k].attrs.get("bounds", None) for k in names])
+            names = set(names) - bounds
+        return names
+
+    def check_results(names, key):
         if scalar_key and len(names) > 1:
             raise KeyError(
-                f"Receive multiple variables for key {k!r}: {names}. "
-                f"Expected only one. Please pass a list [{k!r}] "
-                f"instead to get all variables matching {k!r}."
+                f"Receive multiple variables for key {key!r}: {names}. "
+                f"Expected only one. Please pass a list [{key!r}] "
+                f"instead to get all variables matching {key!r}."
             )
 
     varnames: List[Hashable] = []
@@ -619,6 +629,7 @@ def _getitem(
     for k in key:
         if "coords" not in skip and k in _AXIS_NAMES + _COORD_NAMES:
             names = _get_all(obj, k)
+            names = drop_bounds(names)
             check_results(names, k)
             successful[k] = bool(names)
             coords.extend(names)
@@ -631,6 +642,7 @@ def _getitem(
         else:
             stdnames = set(_get_with_standard_name(obj, k))
             objcoords = set(obj.coords)
+            stdnames = drop_bounds(stdnames)
             if "coords" in skip:
                 stdnames -= objcoords
             check_results(stdnames, k)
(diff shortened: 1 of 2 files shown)


cf_xarray/accessor.py
```diff
@@ -860,7 +871,12 @@ def _get_all_cell_measures(self):
 
         # get all_cell_measures only once
         if not self._all_cell_measures:
-            self._all_cell_measures = set(_CELL_MEASURES + tuple(self.cell_measures))
+            self._all_cell_measures = set(_CELL_MEASURES)
+        try:
+            self._all_cell_measures |= set(self.cell_measures)
+        except ValueError:
+            # ignore bad cell_measures attribute
```
malmans2, line 878: Add a warning here? E.g., we could use the same message of the error.

malmans2 (approved):

cf_xarray/accessor.py
dcherian: how about now? Raise an error with `.cf.cell_measures` but warn in `_getitem`. I think that's better

malmans2: Good idea!

PR #197, opened 2021-04-06
Add unit support to cf-xarray

jthielen:
Closes https://github.com/xarray-contrib/pint-xarray/issues/26

As discussed in https://github.com/xarray-contrib/pint-xarray/issues/26, this PR adds a copy of MetPy's Pint registry (which has some customization to better support CF/UDUNITS-style units most commonly found in CF-compliant NetCDF files) to cf-xarray. However, right now that is *all* that this does at this point, and so, I'd like to put this up now as a work-in-progress and hopefully start a more detailed conversation about how to move forward with this:

- **Main Question**: how should cf-xarray relate to pint-xarray?
  - inherit `.pint` accessor in the `.cf` accessor and use with this registry?
  - set this ported registry to the application registry and provide documentation on how to use it with `pint-xarray`? <-- my preference
  - just provide this ported registry as a standalone entity, but document how to incorporate it and `pint-xarray` into common CF-related workflows?
- Should Pint and this registry be a required or optional dependency?
- How should we handle matplotlib support (which requires enabling on the registry with `units.setup_matplotlib()`) and testing?
- How should this unit support best be documented and tested (in isolation from a calculation suite like MetPy has)?

However, feel free to bring up any other related discussion points as well!

cc @dcherian, @keewis

### DIFF
--- a/cf_xarray/tests/test_units.py
+++ b/cf_xarray/tests/test_units.py
@@ -4,7 +4,11 @@ Reused with modification from MetPy under the terms of the BSD 3-Clause License.
 Copyright (c) 2017 MetPy Developers.
 """
 
+import pytest
+
 from ..units import units
+
+pytest.importskip("pint")
 
 
 def test_added_degrees_units():
--- a/cf_xarray/tests/test_units.py
+++ b/cf_xarray/tests/test_units.py
@@ -49,6 +49,12 @@ def test_percent_units():
     assert str(units("%").units) == "percent"
 
 
+@pytest.mark.xfail(reason="not supported by pint, yet: hgrecco/pint#1295")
 def test_udunits_power_syntax():
     """Test that UDUNITS style powers are properly parsed and interpreted."""
     assert units("m2 s-2").units == units.m ** 2 / units.s ** 2
+
+
+def test_udunits_power_syntax_parse_units():
+    """Test that UDUNITS style powers are properly parsed and interpreted."""
+    assert units.parse_units("m2 s-2") == units.m ** 2 / units.s ** 2
--- a/cf_xarray/tests/test_units.py
+++ b/cf_xarray/tests/test_units.py
@@ -8,7 +8,7 @@ import pytest
 
 from ..units import units
 
-pytest.importskip("pint")
+pytest.importorskip("pint")
 
 
 def test_added_degrees_units():
--- a/cf_xarray/tests/test_units.py
+++ b/cf_xarray/tests/test_units.py
@@ -1,0 +1,50 @@
+r"""Tests the operation of cf_xarray's ported unit support code.
+
+Reused with modification from MetPy under the terms of the BSD 3-Clause License.
+Copyright (c) 2017 MetPy Developers.
+"""
+
+from ..units import units
+
+
+def test_added_degrees_units():
+    """Test that our added degrees units are present in the registry."""
+    # Test equivalence of abbreviations/aliases to our defined names
+    assert str(units("degrees_N").units) == "degrees_north"
+    assert str(units("degreesN").units) == "degrees_north"
+    assert str(units("degree_north").units) == "degrees_north"
+    assert str(units("degree_N").units) == "degrees_north"
+    assert str(units("degreeN").units) == "degrees_north"
+    assert str(units("degrees_E").units) == "degrees_east"
+    assert str(units("degreesE").units) == "degrees_east"
+    assert str(units("degree_east").units) == "degrees_east"
+    assert str(units("degree_E").units) == "degrees_east"
+    assert str(units("degreeE").units) == "degrees_east"
+
+    # Test equivalence of our defined units to base units
+    assert units("degrees_north") == units("degrees")
+    assert units("degrees_north").to_base_units().units == units.radian
+    assert units("degrees_east") == units("degrees")
+    assert units("degrees_east").to_base_units().units == units.radian
+
+
+def test_gpm_unit():
+    """Test that the gpm unit does alias to meters."""
+    x = 1 * units("gpm")
+    assert str(x.units) == "meter"
+
+
+def test_psu_unit():
+    """Test that the psu unit are present in the registry."""
+    x = 1 * units("psu")
+    assert str(x.units) == "practical_salinity_units"
+
+
+def test_percent_units():
+    """Test that percent sign units are properly parsed and interpreted."""
+    assert str(units("%").units) == "percent"
+
+
+def test_udunits_power_syntax():
+    """Test that UDUNITS style powers are properly parsed and interpreted."""
+    assert units("m2 s-2").units == units.m ** 2 / units.s ** 2
--- a/cf_xarray/tests/test_units.py
+++ b/cf_xarray/tests/test_units.py
@@ -41,7 +41,7 @@ def test_gpm_unit():
 def test_psu_unit():
     """Test that the psu unit are present in the registry."""
     x = 1 * units("psu")
-    assert str(x.units) == "practical_salinity_units"
+    assert str(x.units) == "practical_salinity_unit"
 
 
 def test_percent_units():
--- a/cf_xarray/tests/test_units.py
+++ b/cf_xarray/tests/test_units.py
@@ -6,9 +6,9 @@ Copyright (c) 2017 MetPy Developers.
 
 import pytest
 
-from ..units import units
-
 pytest.importorskip("pint")
+
+from ..units import units
 
 
 def test_added_degrees_units():
--- a/cf_xarray/units.py
+++ b/cf_xarray/units.py
@@ -10,7 +10,6 @@ import warnings
 import pint
 from pint import DimensionalityError, UndefinedUnitError, UnitStrippedWarning
 
-
 # Create registry, with preprocessors for UDUNITS-style powers (m2 s-2) and percent signs
 units = pint.UnitRegistry(
     autoconvert_offset_to_baseunit=True,
@@ -22,7 +21,7 @@ units = pint.UnitRegistry(
         ),
         lambda string: string.replace("%", "percent"),
     ],
-    force_ndarray_like=True
+    force_ndarray_like=True,
 )
 
 units.define(
--- a/cf_xarray/units.py
+++ b/cf_xarray/units.py
@@ -1,0 +1,44 @@
+r"""Module to provide unit support via pint approximating UDUNITS/CF.
+
+Reused with modification from MetPy under the terms of the BSD 3-Clause License.
+Copyright (c) 2015,2017,2019 MetPy Developers.
+"""
+import re
+
+import pint
+
+UndefinedUnitError = pint.UndefinedUnitError
+DimensionalityError = pint.DimensionalityError
+UnitStrippedWarning = pint.UnitStrippedWarning
+
+# Create registry, with preprocessors for UDUNITS-style powers (m2 s-2) and percent signs
+units = pint.UnitRegistry(
+    autoconvert_offset_to_baseunit=True,
+    preprocessors=[
+        functools.partial(
+            re.sub,
+            r"(?<=[A-Za-z])(?![A-Za-z])(?<![0-9\-][eE])(?<![0-9\-])(?=[0-9\-])",
+            "**",
+        ),
+        lambda string: string.replace("%", "percent"),
+    ],
+)
+
+units.define(
+    pint.unit.UnitDefinition("percent", "%", (), pint.converters.ScaleConverter(0.01))
+)
+
+# Define commonly encoutered units (both CF and non-CF) not defined by pint
+units.define(
+    "degrees_north = degree = degrees_N = degreesN = degree_north = degree_N = degreeN"
+)
+units.define(
+    "degrees_east = degree = degrees_E = degreesE = degree_east = degree_E = degreeE"
+)
+units.define("@alias meter = gpm")
+units.define("practical_salinity_units = [] = psu")
+
+# Enable pint's built-in matplotlib support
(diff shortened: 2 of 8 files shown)


dcherian:
Thanks @jthielen! This is very exciting to see.

> 2) set this ported registry to the application registry and provide documentation on how to use it with pint-xarray? 
> 3) just provide this ported registry as a standalone entity, but document how to incorporate it and pint-xarray into common CF-related workflows?

For 2) would the registry would be set at import? I think we may want the user to opt in with  `cfxr.setup_udunits_registry()` (or something like that). I'm really torn here :) I guess it's not a bad assumption that anyone importing cf_xarray would be reasonably happy with setting this registry when possible. Curious to hear other opinions here.

> Should Pint and this registry be a required or optional dependency?

Optional. I'd like for cf_xarray to be an easy import that really only depends on xarray for most things.

> How should we handle matplotlib support?

This is very cool! But since the user needs to set units manually on the axes, perhaps we can also expect them to call `ureg.setup_matplotlib(True)` if they want this feature? Also curious to here opinions from people that use this feature.

> How should this unit support best be documented and tested

I think we can test quantifying DataArrays with units like `"degrees_north"` etc. that this registry provides.

dcherian:
> perhaps we can also expect them to call ureg.setup_matplotlib(True) if they want this feature? 

Actually is there any downside to calling `setup_matplotlib` if the user doesn't set units on the axes? What happens then?

keewis (commented):
looks good to me.

The only thing I still have to figure out is how the registry would be set. Using `pint.set_application_registry` doesn't really work because `pint_xarray` sets its default registry on import, which means that we'd have to make sure `pint_xarray` is imported after `cf_xarray` (should be fine if `isort` is used).

cf_xarray/units.py
```diff
@@ -0,0 +1,44 @@
+r"""Module to provide unit support via pint approximating UDUNITS/CF.
+
+Reused with modification from MetPy under the terms of the BSD 3-Clause License.
+Copyright (c) 2015,2017,2019 MetPy Developers.
+"""
+import re
```
keewis, line 6:
```suggestion
import functools
import re
```

cf_xarray/units.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Define commonly encoutered units (both CF and non-CF) not defined by pint
+units.define(
+    "degrees_north = degree = degrees_N = degreesN = degree_north = degree_N = degreeN"
+)
+units.define(
+    "degrees_east = degree = degrees_E = degreesE = degree_east = degree_E = degreeE"
+)
+units.define("@alias meter = gpm")
+units.define("practical_salinity_units = [] = psu")
```
dcherian, line 40:
```suggestion
units.define("practical_salinity_unit = [] = psu")
```

cf_xarray/units.py
```diff
@@ -0,0 +1,44 @@
...
+import re
+
+import pint
+
+UndefinedUnitError = pint.UndefinedUnitError
+DimensionalityError = pint.DimensionalityError
+UnitStrippedWarning = pint.UnitStrippedWarning
+
+# Create registry, with preprocessors for UDUNITS-style powers (m2 s-2) and percent signs
+units = pint.UnitRegistry(
```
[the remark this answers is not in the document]
dcherian: > if this is supposed to work with pint_xarray

:+1:

> optional dependency to pint_xarray is acceptable, you could also pass it through pint_xarray.setup_registry()

I think this is OK. @jthielen?

cf_xarray/tests/test_units.py
```diff
@@ -0,0 +1,50 @@
...
+def test_gpm_unit():
+    """Test that the gpm unit does alias to meters."""
+    x = 1 * units("gpm")
+    assert str(x.units) == "meter"
+
+
+def test_psu_unit():
+    """Test that the psu unit are present in the registry."""
+    x = 1 * units("psu")
+    assert str(x.units) == "practical_salinity_units"
```
dcherian, line 40:
```suggestion
    assert str(x.units) == "practical_salinity_unit"
```

cf_xarray/units.py
```diff
@@ -0,0 +1,45 @@
+r"""Module to provide unit support via pint approximating UDUNITS/CF.
+
+Reused with modification from MetPy under the terms of the BSD 3-Clause License.
```
dcherian, line 3: nice!

cf_xarray/units.py
```diff
@@ -0,0 +1,44 @@
...
+Copyright (c) 2015,2017,2019 MetPy Developers.
+"""
+import re
+
+import pint
+
+UndefinedUnitError = pint.UndefinedUnitError
+DimensionalityError = pint.DimensionalityError
+UnitStrippedWarning = pint.UnitStrippedWarning
```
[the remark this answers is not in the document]
jthielen: Indeed, not sure of the rationale behind why MetPy did it this way (from ~5 yrs ago), but no reason not to adjust it here.

cf_xarray/units.py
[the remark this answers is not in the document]
jthielen: It should be just as easy to set `force_ndarray_like=True`. MetPy's usage predates that option in Pint, so I'm not sure if that breaks anything on MetPy's end. This is a good excuse to test that out, however!

jthielen:
> Actually is there any downside to calling `setup_matplotlib` if the user doesn't set units on the axes? What happens then?

Now that you mention it, no, I don't believe so, as it just sets `matplotlib.units.registry[registry.Quantity]` so that matplotlib can recognize `Quantity` instances as having a special converter instead of a duck array type that gets downcast to ndarray. The only downside I can see to calling `setup_matplotlib` is if a user doesn't have matplotlib installed. For that case, I just wrapped it in an try-except block.

jthielen:
Current todo list remaining:

- [ ] add documentation
  - Examples of using cf-xarray and pint-xarray together
  - Instructing users to either import cf-xarray before pint-xarray or set `pint_xarray.unit_registry = cf_xarray.units.units`
  - List current known shortfalls of this registry compared to full UDUNITS support (namely case sensitivity)
- [ ] add tests of this registry's quantities being used within xarray, and not just correctly parsing units with the customized features.

dcherian:
> I can see to calling setup_matplotlib is if a user doesn't have matplotlib installed. For that case, I just wrapped it in an try-except block.

Perfect.

> Instructing users to either import cf-xarray before pint-xarray or set pint_xarray.unit_registry = units

Is there a reason someone using cf-xarray (requiring xarray) would want the units registry (requiring pint) without using pint_xarray (requires both xarray and pint)? If not, shall we set `pint_xarray.unit_registry = units` in `units.py`?

> it seems like setting it as the application registry is the best solution for most users

I don't know enough to have an opinion here. Up to you, @keewis and @jthielen

dcherian:
Test failure is real.

``` python
=================================== FAILURES ===================================
__________________________ test_udunits_power_syntax ___________________________

    def test_udunits_power_syntax():
        """Test that UDUNITS style powers are properly parsed and interpreted."""
>       assert units("m2 s-2").units == units.m ** 2 / units.s ** 2

cf_xarray/tests/test_units.py:54: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/share/miniconda/envs/cf_xarray_test/lib/python3.7/site-packages/pint/registry.py:1264: in parse_expression
    lambda x: self._eval_token(x, case_sensitive=case_sensitive, **values)
/usr/share/miniconda/envs/cf_xarray_test/lib/python3.7/site-packages/pint/pint_eval.py:93: in evaluate
    return bin_op[op_text](left, self.right.evaluate(define_op, bin_op, un_op))
/usr/share/miniconda/envs/cf_xarray_test/lib/python3.7/site-packages/pint/pint_eval.py:93: in evaluate
    return bin_op[op_text](left, self.right.evaluate(define_op, bin_op, un_op))
/usr/share/miniconda/envs/cf_xarray_test/lib/python3.7/site-packages/pint/quantity.py:115: in wrapped
    return f(self, *args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Quantity(1, 'second')>, other = -2

    @check_implemented
    def __pow__(self, other):
        ...
            new_self = self
            if other == 1:
                return self
            elif other == 0:
                exponent = 0
                units = self.UnitsContainer()
            else:
                if not self._is_multiplicative:
                    if self._REGISTRY.autoconvert_offset_to_baseunit:
                        new_self = self.to_root_units()
                    else:
                        raise OffsetUnitCalculusError(self._units)
    
                if getattr(other, "dimensionless", False):
                    exponent = other.to_root_units().magnitude
                    units = new_self._units ** exponent
                elif not getattr(other, "dimensionless", True):
                    raise DimensionalityError(other._units, "dimensionless")
                else:
                    exponent = _to_magnitude(
                        other, self.force_ndarray, self.force_ndarray_like
                    )
                    units = new_self._units ** exponent
    
>           magnitude = new_self._magnitude ** exponent
E           ValueError: Integers to negative integer powers are not allowed.
```

keewis:
that's a `numpy` issue:
```python
In [3]: np.array(2) ** -1
ValueError                                Traceback (most recent call last)
<ipython-input-3-8b4168950dcb> in <module>
----> 1 np.array(2) ** -1

ValueError: Integers to negative integer powers are not allowed.
```
which can be fixed by casting the exponent to `float`

cf_xarray/tests/test_units.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_percent_units():
+    """Test that percent sign units are properly parsed and interpreted."""
+    assert str(units("%").units) == "percent"
+
+
+def test_udunits_power_syntax():
+    """Test that UDUNITS style powers are properly parsed and interpreted."""
+    assert units("m2 s-2").units == units.m ** 2 / units.s ** 2
```
keewis, line 54: not sure if we need a complete fix, but we could work around this:
```suggestion
    assert units.parse_units("m2 s-2") == units.m ** 2 / unit.s ** 2
```
which would pass and is also what `ds.pint.quantify` uses. Maybe documenting that `units.__call__` and `units.parse_expression` won't work is enough?

cf_xarray/units.py
```diff
@@ -0,0 +1,53 @@
...
+import pint
+from pint import DimensionalityError, UndefinedUnitError, UnitStrippedWarning
+
+# Create registry, with preprocessors for UDUNITS-style powers (m2 s-2) and percent signs
+units = pint.UnitRegistry(
+    autoconvert_offset_to_baseunit=True,
+    preprocessors=[
+        functools.partial(
+            re.sub,
+            r"(?<=[A-Za-z])(?![A-Za-z])(?<![0-9\-][eE])(?<![0-9\-])(?=[0-9\-])",
```
keewis, line 19:
```suggestion
            re.compile(r"(?<=[A-Za-z])(?![A-Za-z])(?<![0-9\-][eE])(?<![0-9\-])(?=[0-9\-])").sub,
```

jthielen:
Sorry for the delayed response here!

> Is there a reason someone using cf-xarray (requiring xarray) would want the units registry (requiring pint) without using pint_xarray (requires both xarray and pint)? If not, shall we set `pint_xarray.unit_registry = units` in `units.py`?

Not that I can see! While we could set `pint_xarray.default_registry = units` in `units.py`, we'd have to make sure the import ordering is addressed properly, since `pint_xarray.default_registry` gets set to the application registry in pint-xarray. Some careful tests with both import orders are likely needed to verify we get the intended behavior.

> not sure if we need a complete fix, but we could work around the failure:
> 
> `...`
>
> which would pass and is also what `ds.pint.quantify` uses. Maybe documenting that `units.__call__` and `units.parse_expression` won't work in general with this type of definition is enough?

I really do intend for that test to pass as-is, since if we're trying to support UDUNITS syntax, I'd think that we don't want to just support it in the units attribute/being parsed with the verbose method, but through the full standard usage of pint; i.e., something like the following should "just work":

```python
new_thing = (2 * units("s-1")) * quantified_data_array
```

If we don't want to hold up this PR for such a fix (such as resolving the scalar issue that necessitates using `force_ndarray_like` in pint-xarray or tweaking pint's parsing internals to better handle this case), however, we can definitely xfail it for now.

keewis:
> tweaking pint's parsing internals

I think that's the way to go: having `parse_expression` fail because `force_ndarray` or `force_ndarray_like` is active seems like a bug in `pint` (we still should investigate whether `xarray` can treat `x.ndim == 0` as a scalar and having `pint.Quantity.ndim` default to `0` if the magnitude raises an `AttributeError` is possible).

What do you think about adding an additional test for `parse_units` and xfailing the `parse_expression` test?

keewis (commented):
@dcherian, with a fix being discussed in hgrecco/pint#1295 I think we should xfail the `__call__` / `parse_expression` test and add a new one for use with `parse_units` (we might be able to remove that new test should `pint` change `__call__` to to call `parse_units`). That would leave only the documentation and examples (I don't think testing `pint` + `xarray` with the new unit registry is really necessary, but I would check that the unit registry is used by `pint-xarray`)

dcherian:
so... is this ready to go in?

keewis:
the code should be ready (if you don't count the integration test mentioned by @jthielen), but the docs are still missing. If you don't mind a separate PR for that I think this can be merged.

dcherian:
Thanks @jthielen & @keewis . Excited to see this go in!

Issue #190, opened 2021-03-18
use `"positive"` attribute in `differentiate` and `integrate`


dcherian:
The `'positive': 'up'` or `'positive': 'down'` attribute should be taken into account when calculating derivatives.

I think we could add `.cf.differentiate(..., fix_sign=True)` (default False) as an opt-in to this approach. In practice, I think we just fix the sign of the coordinate variable before passing it on to xarray.

Thoughts on adding this?

I think we could come up with a better kwarg name.

I think it should raise an error if `fix_sign is True and "positive" not in attrs`

malmans2:
Sounds good to me.

Just one thought about adding kwargs to xarray functions. 
Maybe it's easier if we always use the same kwarg, either a bool (like `use_cf`) or a dict (like `cf_kwargs`)?
It might make more clear what is added by `cf_xarray` and easier to switch on/off all `cf_xarray` extra features.

PR #198, opened 2021-04-07
Add differentiate with positive_upward flag.

dcherian:
xref #190

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1547,7 +1547,7 @@ class CFAccessor:
         """
 
         coord = apply_mapper(
-            (_single(_get_coords),), self._obj, coord, error=False, default=coord
+            (_single(_get_coords),), self._obj, coord, error=False, default=[coord]
         )[0]
         result = self._obj.differentiate(
             coord, edge_order=edge_order, datetime_unit=datetime_unit
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1482,6 +1482,91 @@ class CFAccessor:
             dimensions.update({key: tuple(itertools.chain(*updates))})
         return self._obj.stack(dimensions)
 
+    def differentiate(
+        self,
+        coord,
+        edge_order: int = 1,
+        datetime_unit: str = None,
+        follow_positive: bool = False,
+    ):
+        """ Differentiate the array with the second order accurate central
+        differences.
+
+        .. note::
+            This feature is limited to simple cartesian geometry, i.e. coord
+            must be one dimensional.
+
+        Parameters
+        ----------
+        coord : hashable
+            The coordinate to be used to compute the gradient.
+        edge_order : {1, 2}, default: 1
+            N-th order accurate differences at the boundaries.
+        datetime_unit : {"Y", "M", "W", "D", "h", "m", "s", "ms", \
+                         "us", "ns", "ps", "fs", "as"} or None, optional
+            Unit to compute gradient. Only valid for datetime coordinate.
+        follow_positive: optional, bool
+            Fix sign of the derivative based on the ``"positive"`` attribute of ``coord``.
+            If ``positive=="down"``, then multiplied by -1.
+
+        Returns
+        -------
+        differentiated: DataArray
+
+        See also
+        --------
+        numpy.gradient: corresponding numpy function
+
+        Examples
+        --------
+
+        >>> da = xr.DataArray(
+        ...     np.arange(12).reshape(4, 3),
+        ...     dims=["x", "y"],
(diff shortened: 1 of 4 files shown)


cf_xarray/tests/test_accessor.py
```diff
@@ -1091,3 +1091,43 @@ def test_stack(obj):
...
+    actual = da.cf.differentiate("z", 2, follow_positive=True)
+    assert_identical(expected, actual)
+
+    with da.isel(z=slice(None, None, -1)) as da:
+        expected = -1 * da.differentiate("z", 2)
+        actual = da.cf.differentiate("z", 2, follow_positive=True)
+        assert_identical(expected, actual)
+
+    with xr.set_options(keep_attrs=True):
+        da["z"] = da.z * -1
```
dcherian, line 1117: I think this particular test is right but please check!

cf_xarray/accessor.py
```diff
@@ -1482,6 +1482,91 @@ def stack(self, dimensions=None, **dimensions_kwargs):
             dimensions.update({key: tuple(itertools.chain(*updates))})
         return self._obj.stack(dimensions)
 
+    def differentiate(
+        self,
+        coord,
+        edge_order: int = 1,
+        datetime_unit: str = None,
+        follow_positive: bool = False,
```
dcherian, line 1490: maybe we need a different kwarg name. This option makes it so that positive is always up like in a right-handed coordinate system. Ideas?

cf_xarray/accessor.py
```diff
@@ -1482,6 +1482,91 @@ def stack(self, dimensions=None, **dimensions_kwargs):
...
+        datetime_unit : {"Y", "M", "W", "D", "h", "m", "s", "ms", \
+                         "us", "ns", "ps", "fs", "as"} or None, optional
+            Unit to compute gradient. Only valid for datetime coordinate.
+        follow_positive: optional, bool
+            Fix sign of the derivative based on the ``"positive"`` attribute of ``coord``.
+            If ``positive=="down"``, then multiplied by -1.
+
+        Returns
+        -------
+        differentiated: DataArray
```
malmans2, line 1514: What if we differentiate a Dataset?

Is it feasible to use a decorator to append `self._obj.differentiate` docstring?
Maybe if these kind of methods look like this it's easier to maintain them:
```python
def differentiate(self, *xr_args, follow_positive=False, **xr_kwargs):
    """
    cf_xarray docstring
                +
    appended xarray docstring
    """
```

cf_xarray/tests/test_accessor.py
malmans2: I think it's correct.

cf_xarray/tests/test_accessor.py
```diff
@@ -1091,3 +1091,43 @@ def test_stack(obj):
...
+    actual = da.cf.differentiate("z", 2, follow_positive=True)
+    assert_identical(expected, actual)
+
+    del da.z.attrs["positive"]
+    with pytest.raises(ValueError):
+        da.cf.differentiate("z", follow_positive=True)
+
+    da.z.attrs["positive"] = "zzz"
+    with pytest.raises(ValueError):
+        da.cf.differentiate("z", follow_positive=True)
```
malmans2, line 1133: Should we test a dataset as well and/or `positive = "up"`?

cf_xarray/accessor.py
malmans2: I can't think of any good alternative. Maybe we just document the coordinate system assumed by `cf_xarray`?

cf_xarray/accessor.py
dcherian: > append self._obj.differentiate docstring?

Could not figure this out. So i just simplified the docstring and referred the user to the avtual xarray function.s

malmans2 (approved):

cf_xarray/accessor.py
```diff
@@ -1482,6 +1482,44 @@ def stack(self, dimensions=None, **dimensions_kwargs):
...
+
+        positive_upward: optional, bool
+            Change sign of the derivative based on the ``"positive"`` attribute of ``coord``
+            so that positive values indicate increasing upward.
+            If ``positive=="down"``, then multiplied by -1.
+
+        See Also
+        --------
+        DataArray.cf.differentiate
+        Dataset.cf.differentiate
```
malmans2, line 1502:
```suggestion
        DataArray.cf.differentiate
        Dataset.cf.differentiate
        xarray.DataArray.differentiate: underlying xarray function
        xarray.Dataset.differentiate: underlying xarray function
```

malmans2 (approved):

cf_xarray/accessor.py
malmans2: Sounds good!

PR #200, opened 2021-04-13
Add thickness as special measure name?

dcherian:
I wrote this looking at https://cmip6-preprocessing.readthedocs.io/en/latest/parsing_metrics.html. The thickness variable has `standard_name: cell_thickness` and is called `thkcello` in OMIP/CMIP/CMOR datasets. I *thought* datasets were setting the `cell_measures: "thickness: thkcello"` attribute but now I see that that is not true.

But since I had the code ready, here's the PR. I am not convinced we need to add this.

Also see #201 

cc @jbusecke

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -42,7 +42,7 @@ _AXIS_NAMES = ("X", "Y", "Z", "T")
 _COORD_NAMES = ("longitude", "latitude", "vertical", "time")
 
 #:  Cell measures understood by cf_xarray.
-_CELL_MEASURES = ("area", "volume")
+_CELL_MEASURES = ("thickness", "area", "volume")
 
 # Define the criteria for coordinate matches
 # Copied from metpy
@@ -362,7 +362,7 @@ def _get_with_standard_name(
 def _get_all(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     """
     One or more of ('X', 'Y', 'Z', 'T', 'longitude', 'latitude', 'vertical', 'time',
-    'area', 'volume'), or arbitrary measures, or standard names
+    'thickness', 'area', 'volume'), or arbitrary measures, or standard names
     """
     all_mappers = (_get_axis_coord, _get_measure, _get_with_standard_name)
     results = apply_mapper(all_mappers, obj, key, error=False, default=None)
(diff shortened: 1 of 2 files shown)


jbusecke:
If I understand this correctly (and I really need to dig more into this great package here), this would be very helpful for cmip6_preprocessing. Sorry for asking the stupid question, but this would enable you to write `ds.air.cf.weighted("area").mean(...)` for the datasets that have been parsed here (https://cmip6-preprocessing.readthedocs.io/en/latest/parsing_metrics.html)?

Can I combine metrics too? E.g. `ds.air.cf.weighted(["area", "thickness"])?

jbusecke:
I am thinking how I can parse these cf attributes if they are not present though. Perhaps I could add another input to `parse_metrics(...., {'internal_name_a':'cf_standard_name_a', 'internal_name_b':'cf_standard_name_b'})`, which would map the names I use internally (`areacello`, ...) to cf standard names?

dcherian:
> ds.air.cf.weighted("area").mean(...)

Yes you can already do this. Those variables have `cell_measures: "area: areacello"` set.

> Can I combine metrics too? E.g. ds.air.cf.weighted(["area", "thickness"])?

Nope. 

> I can parse these cf attributes if they are not present though.

We are disucssing autoguessing here: #201. We can hardcode in a heuristic for CMOR names: `areacello` and `thkcello`. In `cmip6_preprocessing` you could set the `cell_measures` attribute using either cf_xarray to autoguess or hardcodeit in.

jbusecke:
This gives me even more reason to agree with you on not making this an xgcm-lite(as mentioned over at #201). I am planning on factoring as much as possible of the renaming/guessing out to cf-xarray, so that change would be very welcome!

dcherian:
> ds.air.cf.weighted(["area", "thickness"])?

You could do `ds.air.cf.weighted("volume")` if you set the `cell_measures: "area: areacello volume: volcello"` and `volcello` is calculated as `areacello * thkcello` (or something appropriate)

jbusecke:
I see, but honestly, it **is** reimplementing what we already did in xgcm. So I think its worth pushing towards that in cmip6_pp.

jbusecke:
This (from #201) is really the reason to go one level up to xgcm, if I understand correctly?

> It is annoying that CF doesn't describe length measures (except for cell_thickness).

dcherian:
> I see, but honestly, it is reimplementing what we already did in xgcm. 

yes. it's just quite lightweight, and a common diagnostic, and we can do it if the attributes are set. 

> This (from #201) is really the reason to go one level up to xgcm

Well that and all the other grid-aware computations. 

It's convenient to do vol/area-weighted averages without having to create a xgcm.Grid object, so I think cf_xarray should provide it. There are natural limits though...

jbusecke:
Fair. Once we make xgcm setup easier this might not be quite true anymore, but I think its not a huge problem to have both ways exist in parallel for now.

PR #212, opened 2021-04-20
Compile regexes

dcherian:
Also fixed a few skipped and xfailed tests


dcherian:
Low risk thanks to tests added by Mattia, so I'm merging.

PR #214, opened 2021-04-21
add bounds property

### LINKED ISSUE #210: Add `cf.bounds`?
Should we add a `bounds` attribute and make `obj.cf` show all bounds?

If yes, how do you think it should look like?
Maybe: `{"bound_lon": "lon", "bound_lat": "lat"}`

I don't think you can have a bound variable associated with multiple variables, so both keys and values can be `str`.

For example, we could use `cf.bounds` to take care of bounds in `rename_like`...

### LINKED ISSUE #215: Revise `cf.standard_names` docstring
https://github.com/xarray-contrib/cf-xarray/blob/64e86954c46642df6f11323d1442fbc70c8fd1bb/cf_xarray/accessor.py#L1145-L1156

We are returning a dict (not a list) for both dataarrays and datasets (not just datasets).

malmans2:
Closes #210
Closes #215 

What should we do with DataArrays, where we can't have bounds because of the extra dimension?

Currently nothing specific is done. For example, when we extract a `DataArray` from a `Dataset`, `da.cf.bounds` is not empty, as we don't pop the `bounds` attribute from the coordinates. However, `da.cf.__repr__` shows `Bounds:  n/a` as the bounds variables are not present.

Should `cf.bounds` only include variables present in the object (and warn about missing bounds)?
Should `cf.bounds` be a property of `Datasets` only?
Should we pop the `bounds` attribute from coordinates when extracting a `DataArray`?
Should we only show `Bounds` in the representation of Datasets?

- [x] Tests
- [x] API docs
- [x] What's new

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -255,6 +255,31 @@ def _get_measure(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     return list(results)
 
 
+def _get_bounds(obj: Union[DataArray, Dataset], key: str) -> List[str]:
+    """
+    Translate from key (either CF key or variable name) to appropriate bounds names.
+    This function interprets the ``bounds`` attribute on DataArrays.
+
+    Parameters
+    ----------
+    obj : DataArray, Dataset
+        DataArray belonging to the coordinate to be checked
+    key : str
+        key to check for.
+
+    Returns
+    -------
+    List[str], Variable name(s) in parent xarray object that matches axis or coordinate `key`
+    """
+
+    results = []
+    for var in apply_mapper(_get_all, obj, key, error=False, default=[key]):
+        if "bounds" in obj[var].attrs:
+            results += [obj[var].attrs["bounds"]]
+
+    return results
+
+
 def _get_with_standard_name(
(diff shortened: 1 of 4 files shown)


dcherian (approved):
Thanks! Looks great

dcherian:
I think we can raise an error for `da.cf.bounds`. 

it's pretty easy to just do `da.attrs["bounds"]` or `da.bounds` to get a string...

cf_xarray/accessor.py
```diff
@@ -461,6 +461,14 @@ def _getattr(
...
         attribute: Union[Mapping, Callable] = getattr(obj, attr)
     except AttributeError:
+        if getattr(
+            CFDatasetAccessor if isinstance(obj, DataArray) else CFDataArrayAccessor,
+            attr,
+            None,
+        ):
+            raise AttributeError(
+                f"{obj.__class__.__name__+'.cf'!r} object has no attribute {attr!r}"
+            )
```
malmans2, line 471: I've added this to try to clarify when `cf_xarray` only adds the attribute to a DataArray/Dataset or the attribute is not present in the underlying `xarray` object.

dcherian: nice.thanks!

PR #221, opened 2021-05-01
`add_bounds` uses `keys` rather than `dims`

malmans2:


cf_xarray/accessor.py
```diff
@@ -1609,12 +1609,16 @@ def add_bounds(self, dims: Union[Hashable, Iterable[Hashable]]):
...
-        else:
-            dimensions = dims
+        if isinstance(keys, str):
+            keys = [keys]
+
+        dimensions = set()
+        for key in keys:
+            dimensions |= set(
+                apply_mapper(_get_dims, self._obj, key, error=False, default=[key])
+            )
```
dcherian, line 1619: A set comprehension is cleaner here


```suggestion
        dimensions = {
                apply_mapper(_get_dims, self._obj, key, error=False, default=[key])
                for key in keys
            )
```

dcherian (commented):
Thanks @malmans2 looks OK, just one minor suggestion. Can you add a whats-new note.

cf_xarray/accessor.py
dcherian: Actually shouldn't the default be `key` instead of `[key]` given that you're doing set subttraction later.

malmans2: `apply_mapper` returns a list of dimension names, shouldn't `default` do the same?

Regarding the set comprehension, this wouldn't work: `{["dim1", "dim2"], ["dim3"]}`.

How about this?
```python
dimensions = {dim for key in keys for dim in apply_mapper(_get_dims, self._obj, key, error=False, default=[key])}
```

Or we could use update:
```python
dimensions = set()
for key in keys:
    dimensions.update(apply_mapper(_get_dims, self._obj, key, error=False, default=[key]))
```

dcherian: I like update

PR #223, opened 2021-05-10
Add pooch to binder environment

malmans2:


malmans2:
Both the documentation and binder environments set `python=3.8`. Do we really need it?

dcherian:
> Do we really need it?

Not really I guess.

Issue #77, opened 2020-08-18
warn at import time if xr.core.options.OPTIONS["keep_attrs"] is False


dcherian:
Maybe say something like 

> "cf_xarray works best when xarray keeps attributes by default. Please run `xr.set_options(keep_attrs=True)`"

This may be annoying though. We should definitely mention this in the docs (#73)

jthielen:
I think it would be good to include the warning, but perhaps give the option to disable it with a cf-xarray option or environment variable if the user wants `keep_attrs` to be `False`.

dcherian:
but you can't set a cf_xarray option without importing it first which would show a warning ;)

to be clear, this warning will only be shown once at import time not on every operation.

jthielen:
> but you can't set a cf_xarray option without importing it first which would show a warning ;)
> 
> to be clear, this warning will only be shown once at import time not on every operation.

Ah, so an environment variable it would have to be!

dcherian:
turns out the default is `"default"` so we could also only raise the warning is `OPTIONS["keep_attrs"] == "default"`. So we wouldn't bug anyone who has explicitly opted out of `keep_attrs`

jthielen:
> turns out the default is `"default"` so we could also only raise the warning is `OPTIONS["keep_attrs"] == "default"`. So we wouldn't bug anyone who has explicitly opted out of `keep_attrs`

Even better!

dcherian:
One (minor) issue with implementing this is that linters will format code to something like

``` python
import cf_xarray  # will raise warning here.
import xarray

xr.set_options(keep_attrs=True)
```

making the warning useless.

PR #227, opened 2021-05-22
Don't apply mappers to DataArrays

### LINKED ISSUE #226: `cf.weighted` doesn't work with DataArrays
Looks like `.cf.weighted` breaks with DataArrays.

```python
weights = xr.ones_like(airds["air"])
weights.attrs = {}
airds.cf.weighted(weights)
```
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
Note that I can't do this, because only `Datasets` and `DataArrays` have the `.cf` accessor:
```python
airds.weighted(weights).cf.mean("T")
```
```
AttributeError: 'DatasetWeighted' object has no attribute 'cf'
```

One possible workaround is:
```python
airds.weighted(weights).mean(airds.cf.axes["T"])
```
or
```python
airds = airds.assign_coords(weights=weights)
airds["air"].attrs["cell_measures"] += " weights: weights"
airds.cf.weighted("weights").mean("T")
```
but I don't think that's what we want.

malmans2:
Closes #226

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -90,8 +90,7 @@ def apply_mapper(
             raise ValueError("`default` must be provided when `key` is not a string.")
         return list(always_iterable(default))
 
-    if default is None:
-        default = []
+    default = [] if default is None else list(always_iterable(default))
 
     def _apply_single_mapper(mapper):
 
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -857,11 +857,20 @@ class CFAccessor:
             elif sig.parameters[param].kind is inspect.Parameter.VAR_POSITIONAL:
                 var_args.append(param)
 
-        posargs = []
+        # Catch DataArrays and return them as they are
+        args = list(args)
+        arguments = {}
+        posargs = [
+            args.pop(i) for i, arg in enumerate(args) if isinstance(arg, DataArray)
+        ]
+        for key, value in dict(kwargs).items():
+            if isinstance(value, DataArray):
+                arguments[key] = kwargs.pop(key)
+
         if args or kwargs:
             bound = sig.bind(*args, **kwargs)
-            arguments = self._rewrite_values(
-                bound.arguments, key_mappers, tuple(var_kws)
+            arguments.update(
+                **self._rewrite_values(bound.arguments, key_mappers, tuple(var_kws))
             )
 
             # unwrap the *args type arguments
@@ -875,8 +884,6 @@ class CFAccessor:
                 value = arguments.pop(kw, None)
                 if value:
                     arguments.update(**value)
-        else:
-            arguments = {}
 
         return posargs, arguments
 
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -26,7 +26,12 @@ from xarray.core.arithmetic import SupportsArithmetic
 
 from .criteria import coordinate_criteria, regex
 from .helpers import bounds_to_vertices
-from .utils import _is_datetime_like, invert_mappings, parse_cell_methods_attr
+from .utils import (
+    _is_datetime_like,
+    always_iterable,
+    invert_mappings,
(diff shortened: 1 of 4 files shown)


doc/examples/introduction.ipynb
```diff
@@ -25,6 +25,22 @@
...
+    "`cf_xarray` works best when `xarray` keeps attributes by default.\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "xr.set_options(keep_attrs=True)"
```
malmans2, line 41: Looks like we've been running the intro notebook without `xr.set_options(keep_attrs=True)`.
I don't think the issue without it was triggered by this PR, but I might be wrong.

Anyways, I think we should keep attributes and maybe re-discuss #77?

cf_xarray/accessor.py
```diff
@@ -857,11 +857,20 @@ def _process_signature(
...
+        for key, value in dict(kwargs).items():
+            if isinstance(value, DataArray):
+                arguments[key] = kwargs.pop(key)
+
         if args or kwargs:
             bound = sig.bind(*args, **kwargs)
-            arguments = self._rewrite_values(
-                bound.arguments, key_mappers, tuple(var_kws)
+            arguments.update(
+                **self._rewrite_values(bound.arguments, key_mappers, tuple(var_kws))
```
dcherian, line 873: i'm wondering if it would be cleaner to modify `_rewrite_values` instead.

malmans2: Right! Looks like adding an isinstance check in `_rewrite_values` and `apply_mapper` works well...

cf_xarray/accessor.py
```diff
@@ -79,6 +79,10 @@ def apply_mapper(
     It should return a list in all other cases including when there are no
     results for a good key.
     """
+
+    if not isinstance(key, str) and default is not None:
+        return default
```
malmans2, line 84: Should we add and `error` if key is not a string and default is not None? Although apply mapper is mainly used internally, so the error wouldn't mean much to users?

cf_xarray/accessor.py
```diff
@@ -924,7 +921,9 @@ def _rewrite_values(
             value = kwargs[key]
             mappers = all_mappers[key]
 
-            if isinstance(value, str):
+            if isinstance(value, (str, DataArray, Dataset)) or not isinstance(
+                value, Iterable
+            ):
```
malmans2, line 926: I made it more general, not sure if it's the best thing to do... (e.g., I don't think currently there are wrapped methods where you can pass a `Dataset`).

cf_xarray/accessor.py
dcherian: Yes let's raise an error, it will help when it's raised :)

Something like this?

``` python
if not isinstance(key, str):
    if default is None
        raise
    else:
		return default
```

cf_xarray/accessor.py
dcherian: What if we used this instead... (from https://github.com/pydata/xarray/pull/4972)
``` python
def always_sequence(obj):
    if not isinstance(obj, (list, tuple)):
        obj = [obj]

    return obj
```

malmans2: Should we also add set and dict? Otherwise we are going to return `[{}]`.
Should we place it in `utils.py`?

malmans2: Maybe:
```python
def always_sequence(obj):
    if isinstance(obj, set):
        obj = list(obj)
    elif not isinstance(obj, (list, tuple, dict)):
        obj = [obj]

    return obj
```

dcherian: Yeah that looks right

dcherian: We could call it `always_iterable`

dcherian:
Thanks @malmans2 let's try it!

PR #232, opened 2021-06-08
Implement `.cf.add_canonical_attributes`

malmans2:
- [x] Add tests
- [x] Might be worth adding this feature in the intro notebook
- [x] Add to API doc

Right now by default we're parsing this url: https://raw.githubusercontent.com/cf-convention/cf-convention.github.io/master/Data/cf-standard-names/current/src/cf-standard-name-table.xml

Should we add a copy in cf-xarray?


malmans2:
parser

malmans2:
Should we prevent adding units if variable is `datetime64` or `timedelta64`?
We could use: https://pandas.pydata.org/pandas-docs/version/0.23/generated/pandas.api.types.is_datetime64_dtype.html

aulemahal:
Nice feature!
If I may, I would vote for having the table local : 1. for the few cases with no internet access, 2. to ensure a good correspondence between cf-xarray's version and cf. So we can say, for example,  cf-xarray~=0.5 goes with CF 1.8, and so on.

However, I'm not sure how we can safely attribute units? Correct me if I'm wrong, but in your case above the `time` axis had no units _because_ xarray had parsed it into a datetime dtype and thus it would be somewhat incorrect to say they are "seconds"? I'd say a missing unit attribute on data is a problem from the user side, cf-xarray doesn't have enough info to patch that.

malmans2:
You're right!
There's a `skip` argument so `add_cf_attributes(skip=["units"])` would not add units. Maybe add a note and/or default to skip units?

I'm thinking we can work around this though. Other than time variables, I can't think of other use-cases where `add_cf_attributes(override=False)` would obviously add something wrong.
What do you think about something like:
```python
if hasattr(da, "dt"):
    # Do not add units attribute
```

aulemahal:
Hum, I believe there still could be broken datasets with a missing "unit" attribute? There setting a faulty one would be an error, no?

Agreed for the 'dt' workaround and would go with defaulting to `skip=['units']`. But those edgecases I am talking about shouldn't be very frequent, so I could be convinced otherwise.

cf_xarray/accessor.py
```diff
@@ -1523,6 +1531,69 @@ def differentiate(
                 result *= -1
         return result
 
+    def add_cf_attributes(
```
dcherian, line 1534:
```suggestion
    def add_canonical_cf_attributes(
```

I am 50/50 on this suggestion

dcherian, line 1534:
```suggestion
    def add_canonical_cf_attributes(
```

I'm 50/50 on this suggestion

cf_xarray/utils.py
```diff
@@ -57,3 +61,57 @@ def invert_mappings(*mappings):
 
 def always_iterable(obj: Any) -> Iterable:
     return [obj] if not isinstance(obj, (tuple, list, set, dict)) else obj
+
+
+def parse_cf_table(uri: Union[str, Path], verbose=False):
```
dcherian, line 66: Should we parse this once at release-time and bundle a pickled dictionary? Or write a valid python file that creates the dictionary?

dcherian:
Nice idea!

dcherian:
See `utils._is_datetime_like` for detecting time variables.

One suggestion from the EarthCube review was to add a `history` attribute saying these attributes were added. I think we should do that and mention which version of the standard name table was used.

malmans2:
I'm leaning towards renaming it to `add_canonical_cf_attributes` but add all missing attributes by default (with exception of units for `datetime_like` variables). I think the new name makes more clear what it does, and `skip` allows to pick only attributes needed.

For the actual table, yes let's store it in this repo and use pytest to make sure that we are using the latest version. Maybe json is better than pickle so it's easier to track changes?

aulemahal:
My two cents : in xclim we use yaml files more and more because they are easy to read both as a human and as a python script (and in a git diff!). With a bit more security when using pyyaml's `safe_load`.

malmans2:
Mmm, maybe it's a good idea to only have python files in `cf_xarray/` and complitely skip the parsing part?
`black` actually makes the dictionaries quite readable...

I'm kind of 50/50. `yaml` seems like the best option for this kind of data, but works fine with just a `.py` file.
Also possible downside of using a `yaml`:
1. I believe `pyaml` is not a Python system library?
2. I we use something that is not python, shouldn't we just use the native `xml` file?

dcherian:
> I we use something that is not python, shouldn't we just use the native xml file?

I was worried about speed initially. But maybe this is the simplest option and we should use it till it becomes an issue

malmans2:
Having everything in `.py` seems like a good idea to me. How about raw xml as a str variable in a `.py`?

cf_xarray/accessor.py
```diff
@@ -1523,6 +1524,112 @@ def differentiate(
...
+                        # Build verbose dictionary
+                        attrs_to_print.setdefault(var_name, {})
+                        attrs_to_print[var_name][key] = value
+
+                        # Update history
+                        history = ds[var_name].attrs.get("cf_history", "")
+                        if history:
+                            history += "; "
+                        history += f"add canonical CF attribute {key!r} (v{info['version_number']!s})"
+                        ds[var_name].attrs["cf_history"] = history
```
malmans2, line 1613: Thoughts?
It's kind of quick and dirty, I guess we will eventually have a more robust approach for all methods?
What do you think about using a separate `cf_history` rather than `history`?

dcherian:
> How about raw xml as a str variable in a .py?

I prefer just using the xml file then.

cf_xarray/accessor.py
dcherian: `history` is an NCO convention I think. We'll want to stick this new statement at the front of a history attribute, if any

malmans2:
> > How about raw xml as a str variable in a .py?
> 
> I prefer just using the xml file then.

OK. I think we just need to add a [MANIFEST.in](https://packaging.python.org/guides/using-manifest-in/).

cf_xarray/accessor.py
malmans2: If we add to `history` we should probably align with existing conventions:
1. I think NCO only adds global attributes, should we do that with Datasets?
2. To align with NCO I guess we'd have to add `date: .cf.add_canonical_cf_attributes(arg1=..., arg2=...)` and a separate attribute with cf-xarray and cf-table version?
```
:history = "Wed Feb 24 15:32:09 2021: ncks --no_abc --cnk_dmn x,64 --cnk_dmn y,64 --cnk_dmn nav_lev,1 --cnk_dmn time_counter,1 --dfl_lvl 1 GSRIDGE36Z_mesh_mask.nc GSRIDGE36Z_mesh_mask.nc" ;
:NCO = "netCDF Operators version 4.7.6 (Homepage = http://nco.sf.net, Code = http://github.com/nco/nco)" ;
```

cf_xarray/accessor.py
```diff
@@ -1632,7 +1608,20 @@ def add_canonical_cf_attributes(
...
+        method_name = inspect.stack()[0][3]
+        version = _get_version()
+        table_version = info["version_number"]
+        history = (
+            f"{now}:"
+            f" cf.{method_name}({args})"
+            f" [cf-xarray {version}, cf-standard-name-table {table_version}]\n"
+        )
+        obj = self._maybe_to_dataarray(ds)
+        obj.attrs["history"] = history + obj.attrs.get("history", "")
```
malmans2, line 1622: I guess in the long run we'll need to introduce some kind of decorator...

cf_xarray/tests/test_scripts.py
```diff
@@ -1,32 +1,26 @@
 import os
+from tempfile import TemporaryDirectory
```
malmans2, line 2: Sorry, this change is not really relevant to this PR but I noticed it and I think this is more robust...

malmans2:
OK, this is ready for review... I think naming it `cf.add_canonical_attributes` is clear enough...?

cf_xarray/tests/test_accessor.py
```diff
@@ -417,7 +419,7 @@ def test_args_methods(obj):
 
 def test_dataarray_getitem():
 
-    air = airds.air
+    air = airds.air.copy()
```
malmans2, line 422: Unrelated to this PR, but this was changing the "global" `airds`

cf_xarray/accessor.py
```diff
@@ -1523,6 +1525,104 @@ def differentiate(
...
+    def add_canonical_attributes(
+        self,
+        override: bool = False,
+        skip: Union[str, List[str]] = None,
+        verbose: bool = False,
+        source=None,
+    ) -> Union[Dataset, DataArray]:
+        """
+        Add canonical CF attributes to variables with standard names.
+        Attributes are parsed from the official CF standard name table.
```
dcherian, line 1537: Should link to http://cfconventions.org/standard-names.html and mention that an entry will added to the `history` attribute

cf_xarray/accessor.py
```diff
@@ -1523,6 +1525,104 @@ def differentiate(
...
+                        # Don't add units to time variables (e.g., datetime64, ...)
+                        if key == "units" and _is_datetime_like(ds[var_name]):
+                            continue
+
+                        # Add attribute
+                        ds[var_name].attrs[key] = value
+
+                        # Build verbose dictionary
+                        attrs_to_print.setdefault(var_name, {})
+                        attrs_to_print[var_name][key] = value
```
dcherian, line 1593:
```suggestion
                        if verbose:
                        	attrs_to_print.setdefault(var_name, {})
                        	attrs_to_print[var_name][key] = value
```

cf_xarray/tests/test_accessor.py
```diff
@@ -1233,3 +1235,67 @@ def test_cmip6_attrs():
     )
     assert da.cf.axes["X"] == ["nlon"]
     assert da.cf.axes["Y"] == ["nlat"]
+
+
+def test_cf_standard_name_table_version():
+
+    url = (
+        "https://raw.githubusercontent.com/cf-convention/cf-convention.github.io/"
+        "master/Data/cf-standard-names/current/src/cf-standard-name-table.xml"
```
dcherian, line 1244: is this a released version or a development version?

cf_xarray/tests/test_accessor.py
```diff
@@ -1233,3 +1235,67 @@ def test_cmip6_attrs():
...
+    )
+    expected_info, _, _ = parse_cf_standard_name_table(urlopen(url))
+    actual_info, _, _ = parse_cf_standard_name_table()
+    assert expected_info == actual_info
+
+
+@pytest.mark.parametrize("override", [True, False])
+@pytest.mark.parametrize("skip", ["units", None])
+@pytest.mark.parametrize("verbose", [True, False])
+def test_add_canonical_attributes(override, skip, verbose, capsys):
```
dcherian, line 1254: nice test!

cf_xarray/tests/test_scripts.py
```diff
@@ -1,32 +1,26 @@
...
+        "all_criteria",
+        "all_regex",
+    ]
+    tables_to_check = [f"_build/csv/{name}.csv" for name in names]
 
-        make_doc.main()
-        assert all(os.path.exists(path) for path in tables_to_check)
-    finally:
-        os.chdir(owd)
+    # Create _buil/csv in a temporary directory
```
dcherian, line 17:
```suggestion
    # Create _build/csv in a temporary directory
```

cf_xarray/utils.py
```diff
@@ -57,3 +60,41 @@ def invert_mappings(*mappings):
 
 def always_iterable(obj: Any) -> Iterable:
     return [obj] if not isinstance(obj, (tuple, list, set, dict)) else obj
+
+
+def parse_cf_standard_name_table(source=None):
```
dcherian, line 65: didn't review but if things work, LGTM!

dcherian (approved):
Thanks Mattia! This is a nice change.

I'll leave it open for a review by @aulemahal if he has time.

aulemahal (approved):
I like this, quite clean! Thanks.

dcherian:
V.cool!

One last q: should the default for `skip` be `skip=("grib", "amip")`. This is really a personal preference so no strong feelings.

malmans2:
> V.cool!
> 
> One last q: should the default for `skip` be `skip=("grib", "amip")`. This is really a personal preference so no strong feelings.

I thought about it, but I also prefer having `None` as default for optional arguments. 
I also like being able to do this:
```python
ds.cf.add_canonical_attributes(verbose=True)
```
ooooh, I don't need `amip`:
```python
canonical_ds = ds.cf.add_canonical_attributes(skip="amip")
```

One thing we could change is making attributes case insensitive?
For example, I thing `airds` has the attribute `GRIB`, so with `override=False` we could skip that...
I couldn't figure out if uppercase `GRIB` is an actual convention or not...

dcherian:
> One thing we could change is making attributes case insensitive?

nah, i think we just add lowercase attributes and users can figure it out...

cf_xarray/tests/test_accessor.py
dcherian: It seems like this is the latest stable version

cf_xarray/accessor.py
```diff
@@ -1523,6 +1525,105 @@ def differentiate(
...
+    def add_canonical_attributes(
+        self,
+        override: bool = False,
+        skip: Union[str, List[str]] = None,
+        verbose: bool = False,
+        source=None,
+    ) -> Union[Dataset, DataArray]:
+        """
+        Add canonical CF attributes to variables with standard names.
+        Attributes are parsed from the official CF standard name table.
```
dcherian, line 1537:
```suggestion
        Attributes are parsed from the official CF standard name table.
        This function adds an entry to the "history" attribute.
```

PR #233, opened 2021-06-09
Change coord criteria to {cf_name: criteria}

kthyng:
These changes are in support of updating the `_get_axis_coord` logic to be a bit more generalized so the same logic can be used to find axis coordinates and to pull out variables (that is a future step).

The coordinate_criteria dictionary in criteria.py is now inverted so the keys are axes and coord names with values of valid attribute types and values.

The function `_get_axis_coord()` was updated to try to account for this change in the `coordinate_criteria` dictionary though it may not be correct quite yet. The tests did not all pass.

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -221,8 +221,7 @@ def _get_axis_coord(var: Union[DataArray, Dataset], key: str) -> List[str]:
     results: Set = set()
     for coord in search_in:
         if key in coordinate_criteria:
-            for criterion in coordinate_criteria[key]:
-                expected = coordinate_criteria[key][criterion]
+            for criterion, expected in coordinate_criteria[key].items():
                 if (
                     coord in var.coords
                     and var.coords[coord].attrs.get(criterion, None) in expected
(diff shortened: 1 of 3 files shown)


dcherian:
Looks like you'll have to fix the `make_*` functions in 

https://github.com/xarray-contrib/cf-xarray/blob/e11cf07f49ab121633131a024843644045304505/cf_xarray/scripts/make_doc.py#L20

But everyhting works otherwise!

cf_xarray/accessor.py
```diff
@@ -220,15 +220,14 @@ def _get_axis_coord(var: Union[DataArray, Dataset], key: str) -> List[str]:
 
     results: Set = set()
     for coord in search_in:
-        for criterion, valid_values in coordinate_criteria.items():
-            if key in valid_values:
-                expected = valid_values[key]
+        if key in coordinate_criteria:
+            for criterion in coordinate_criteria[key]:
+                expected = coordinate_criteria[key][criterion]
```
dcherian, line 225:
```suggestion
            for criterion, expected in coordinate_criteria[key].items():
```

dcherian:
Thanks @kthyng this looks great!

PR #234, opened 2021-06-10
Support custom vocabularies/criteria

kthyng:
Working on adding:
1. a way to set_options for user-defined custom_criteria dictionary
2. generalizing `_get_axis_coord` to `get_custom_criteria`
3. Adding a `cf_criteria` dictionary too
4. testing custom dictionary


cf_xarray/criteria.py
```diff
@@ -8,6 +8,34 @@
 import re
 from typing import MutableMapping, Tuple
 
+# KMT: Not finished, but as draft
```
dcherian, line 11: This is not needed. We can create one later for standard_name aliases Otherwise, I intend to leave this up to a user to create.

cf_xarray/tests/test_accessor.py
```diff
@@ -1233,3 +1233,53 @@ def test_cmip6_attrs():
...
+
+def test_custom_criteria():
+    my_custom_criteria = {
+        "ssh": {
+            "standard_name": (
+                "sea_surface_height_amplitude_due_to_geocentric_ocean_tide_geoid_mllw",
+                "sea_surface_height_above_sea_level_geoid_mllw",
+            ),
+            "name": ("sea_surface_elevation",),  # variable name
+            "units": ("m",),
```
dcherian, line 1246: I don't see how `units` could ever work here. With the current logic, everything with units of "m" will match `ds.cf["ssh"]`. That's not right.

cf_xarray/accessor.py
```diff
@@ -201,6 +239,10 @@ def _get_axis_coord(var: Union[DataArray, Dataset], key: str) -> List[str]:
     MetPy's parse_cf
     """
 
+    # KMT: not sure how to set this yet
+    # if criteria is None:
+    #     criteria = OPTIONS["criteria"]  # set by set_options
+
```
dcherian, line 245:
```suggestion
```

cf_xarray/accessor.py
```diff
@@ -168,6 +177,31 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+    if criteria is not None:
+        criteria = always_iterable(criteria, allowed=(tuple, dict, set))
+
+    results: Set = set()
+    for crit in criteria:
+        if key in crit:
+            for criterion, expected in crit[key].items():
+                for var in obj.variables:
+                    if obj[var].attrs.get(criterion, None) in expected:
```
dcherian, line 200: will have to deal with regexes here. Maybe always treat `expected` as a regex and pass to `re.match`?

Also this will need to deal with `"name"` in the criteria. So something like 
``` python
if criterion == 'name' and var == expected:
	results.update((var,))
```

cf_xarray/accessor.py
```diff
@@ -312,7 +354,12 @@ def _get_all(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     One or more of ('X', 'Y', 'Z', 'T', 'longitude', 'latitude', 'vertical', 'time',
     'area', 'volume'), or arbitrary measures, or standard names
     """
-    all_mappers = (_get_axis_coord, _get_measure, _get_with_standard_name)
+    all_mappers = (
+        _get_custom_criteria,
+        _get_axis_coord,
```
dcherian, line 359: I kept `_get_axis_coord` separate for now to keep test passing. We can do some refactoring and cleanup once we get all the logic working.

dcherian (commented):
I pushed some changes. This is exciting!

I think it's ready for some actual test cases.

cf_xarray/accessor.py
```diff
@@ -178,6 +212,10 @@ def _get_axis_coord(var: Union[DataArray, Dataset], key: str) -> List[str]:
         DataArray belonging to the coordinate to be checked
     key : str, ["X", "Y", "Z", "T", "longitude", "latitude", "vertical", "time"]
         key to check for.
+    criteria : dict
```
dcherian, line 215: guess this should be moved.

cf_xarray/accessor.py
```diff
@@ -168,6 +177,31 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+    if isinstance(obj, DataArray):
+        obj = obj._to_temp_dataset()
+
+    if criteria is None:
+        if not OPTIONS["custom_criteria"]:
+            return []
+        criteria = OPTIONS["custom_criteria"]
+
+    if criteria is not None:
+        criteria = always_iterable(criteria, allowed=(tuple, dict, set))
```
dcherian, line 193:
```suggestion
        criteria = always_iterable(criteria, allowed=(tuple, list, set))
```

cf_xarray/criteria.py
kthyng: Ok I removed that.

cf_xarray/tests/test_accessor.py
kthyng: True, removed units.

kthyng (commented):
Ok looks good. I'm not used to navigating the github interactive reviewing like this so I may have missed something.

cf_xarray/accessor.py
kthyng: Ok there are a couple more of your suggestions here that I will look at locally.

kthyng: Can you provide a brief example of how you imagine this working that I can test with locally? I'm struggling with making these suggestions concrete when I don't have an example to run and I am not sure how to come up with it on my own without the vision you have.

dcherian: Using your test dataset, we want this to work

```python
    my_custom_criteria = {
        "ssh": {
            "standard_name": (
                "sea_surface_height_*",
            ),
            "name": ("sea_surface_elevation",),  # variable name
        }
	}

	cfxr.set_options(my_custom_criteria)

	ds = xr.Dataset()
	ds["elev"] = ("dim", np.arange(10), {"standard_name": "sea_surface_height_asd"})
	ds["sea_surface_elevation"] = ("dim", np.arange(10))

	xr.testing.assert_identical(ds.cf["ssh"], ds["elev"])
	xr.testing.assert_identical(ds.cf["ssh"], ds["sea_surface_elevation"])
```

Does this look like what you want?

kthyng: Yes, I think that is a good example.

cf_xarray/accessor.py
kthyng: Sounds good.

kthyng:
@dcherian Got it to work with your example such that the following work:
```
cfxr.accessor._get_custom_criteria(ds, "ssh")
```
to return `['elev']` and 
```
cfxr.accessor._get_all(ds, 'ssh')
``` 
to return `['elev']`, but not sure how to connect to make `ds.cf["ssh"]` work.

dcherian:
Nice, I think we need to update `_getitem` too: https://github.com/xarray-contrib/cf-xarray/blob/2fa2939181e922ee872bc2241367dbc48d07c6dc/cf_xarray/accessor.py#L589-L614

kthyng:
> Nice, I think we need to update `_getitem` too:

Ok I made some changes. I am including the custom variable mapping in `varnames` — is that ok? Looks like it works anyway.

cf_xarray/accessor.py
```diff
@@ -584,6 +663,11 @@ def check_results(names, key):
         measures = []
         warnings.warn("Ignoring bad cell_measures attribute.", UserWarning)
 
+    try:
+        custom_criteria = OPTIONS["custom_criteria"][0]
+    except IndexError:
+        custom_criteria = OPTIONS["custom_criteria"]
```
dcherian, line 669: I don't see why we need `IndexError`. `OPTIONS['custom_criteria']` should always be a list.

We can use the `ChainMap` trick here too to go through all criteria.

cf_xarray/accessor.py
```diff
@@ -600,6 +684,11 @@ def check_results(names, key):
             successful[k] = bool(measure)
             if measure:
                 varnames.extend(measure)
+        elif k in custom_criteria:
```
dcherian, line 687: This looks fine. We should try to refactor this logic, maybe by pushing the `skip` argument down to `_get_all`. cc @malmans2

cf_xarray/tests/test_accessor.py
```diff
@@ -1233,3 +1233,34 @@ def test_cmip6_attrs():
...
+            "standard_name": ("northward_sea_water_velocity",),
+        },
+        "wind_speed": {
+            "standard_name": ("wind_speed",),
+        },
+        "wind_speed_of_gust": {
+            "standard_name": ("wind_speed_of_gust",),
+        },
+    }
+    cf_xarray.accessor.set_options(my_custom_criteria)
```
dcherian, line 1266: I think if you add your tests here, we should be done.

dcherian (commented):
What do you think of calling this "custom criteria". SHould we call it "custom vocabulary" instead?

cf_xarray/accessor.py
```diff
@@ -168,6 +177,71 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
         return []
 
 
+def _get_custom_criteria(
```
dcherian, line 180: The docstring needs to be updated

cf_xarray/accessor.py
```diff
@@ -168,6 +177,71 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+    if criteria is None:
+        if not OPTIONS["custom_criteria"]:
+            return []
+        criteria = OPTIONS["custom_criteria"]
+
+    if criteria is not None:
+        criteria = always_iterable(criteria, allowed=(tuple, list, set))
+
+    results: Set = set()
+    for crit in criteria:
```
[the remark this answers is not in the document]
kthyng: I integrated ChainMap in https://github.com/xarray-contrib/cf-xarray/pull/234/commits/0125789e332202755ea6c7ffcbd9999e6dfbb8ab

I like custom_criteria since it matches the criteria sort of format in `criteria.py`.

cf_xarray/accessor.py
kthyng: I didn't do anything on this since not sure about the `skip` logic.

cf_xarray/accessor.py
kthyng: I made some updates in https://github.com/xarray-contrib/cf-xarray/pull/234/commits/0125789e332202755ea6c7ffcbd9999e6dfbb8ab

cf_xarray/tests/test_accessor.py
kthyng: I will work on tests now.

cf_xarray/accessor.py
```diff
@@ -168,6 +177,73 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+    criteria : dict
+        Criteria to use to map from variable to attributes describing the
+        variable. An example is coordinate_criteria which maps coordinates to
+        their attributes and attribute values. If user has defined
+        custom_criteria, this will be used by default.
+    error : bool
+        raise errors when key is not found or interpretable. Use False and provide default
+        to replicate dict.get(k, None).
+    default : Any
+        default value to return when error is False.
```
dcherian, line 201:
```suggestion
```

cf_xarray/accessor.py
dcherian: yes. sorry I didn't make clear that that was a suggestion for the future

kthyng:
As I am working on tests, I also want to see how it will integrate into the other packages that need this sort of behavior. I realize now that another thing I need is to be able to not just ask for the "ssh" variable for example, but ask which type of variable I have. So in addition to:
```
ds.cf["ssh"]
```
returning the "ssh"-matching variable, I would like to be able to, for a DataArray da:
```
da.cf._get_custom_type()
```
to return "ssh", or somehow do the inverse search. Any ideas how you would like to see this happen @dcherian?

kthyng:
@dcherian I am planning to add more tests but am out of time for today so will add more tomorrow.

cf_xarray/accessor.py
```diff
@@ -168,6 +177,68 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+    This functions checks for the following attributes in order for axis
+    or coordinate:
+       - `standard_name` (CF option)
+       - `_CoordinateAxisType` (from THREDDS)
+       - `axis` (CF option)
+       - `positive` (CF standard for non-pressure vertical coordinate)
+
+    References
+    ----------
+    MetPy's parse_cf
```
dcherian, line 213:
```suggestion
```

cf_xarray/accessor.py
```diff
@@ -168,6 +177,68 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+    """
+    Translate from axis, coord, or custom name to variable name
+
+    Parameters
+    ----------
+    obj : DataArray, Dataset
+        DataArray belonging to the coordinate to be checked
+    key : str, ["X", "Y", "Z", "T", "longitude", "latitude", "vertical", "time"], or a key from user-defined custom_criteria
+        key to check for.
+    criteria : dict
```
dcherian, line 192:
```suggestion
    criteria : dict, optional
```

cf_xarray/accessor.py
```diff
@@ -168,6 +177,68 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
         return []
 
 
+def _get_custom_criteria(
+    obj: Union[DataArray, Dataset], key: str, criteria=None
+) -> List[str]:
+    """
+    Translate from axis, coord, or custom name to variable name
```
dcherian, line 184:
```suggestion
    Translate from axis, coord, or custom name to variable name optionally
    using ``custom_criteria``
```

dcherian:
oh hmm...

Do you want to know type or do you want to assert that it is a specific type?

```
assert da.cf.matches_custom_criteria(key="ssh")
```

kthyng:
> Do you want to know type or do you want to assert that it is a specific type?

I would prefer the flexibility of knowing the type (not asserting).

kthyng:
@dcherian Should we wrap up this PR instead of trying to add the additional functionality of returning the variable name (like being able to query a DataArray to return "ssh" in my custom_criteria), or is it easy enough to add to wait?

cf_xarray/accessor.py
```diff
@@ -168,6 +177,57 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+
+    results: Set = set()
+    if key in criteria:
+        for criterion, expected in criteria[key].items():
+            for var in obj.variables:
+                # Treat expected as regex
+                if re.match("|".join(expected), obj[var].attrs.get(criterion, "")):
+                    results.update((var,))
+                # also check name specifically since not in attributes
+                elif criterion == "name" and var in expected:
```
dcherian, line 226: I bet someone will want the regex version here too.

cf_xarray/tests/test_accessor.py
```diff
@@ -1233,3 +1233,23 @@ def test_cmip6_attrs():
...
+
+    # Match by standard_name regex match
+    ds = xr.Dataset()
+    ds["elev"] = ("dim", np.arange(10), {"standard_name": "sea_surface_elevBLAH"})
+    xr.testing.assert_identical(ds.cf["ssh"], ds["elev"])
+
+    # Match by exact name
+    ds = xr.Dataset()
+    ds["sea_surface_elevation"] = ("dim", np.arange(10))
+    xr.testing.assert_identical(ds.cf["ssh"], ds["sea_surface_elevation"])
```
dcherian, line 1255: can we adda  second ssh-like variable and test that we get back a dataset for `ds.cf[["ssh"]]` and that it raises and error for `ds.cf["ssh"]`

dcherian:
:+1: for small functional PRs! I just have a couple of test requests. Also there are some merge conflicts. It also needs a whats-new entry

cf_xarray/accessor.py
kthyng: Do you mean someone would want regex in the name check? I removed it because I thought maybe for "name" comparison, people would want it to be exact. The problem for me that prompted that was I had variables "wind_speed" and "wind_speed_of_gust", and the check for the first was picking up the second. Or is there a better way to separate?

kthyng:
What's the best way to go through the merge conflicts?

dcherian:
THe github interface works OK for small changes (usually you'll end up with a formatting error).

Otherwise pull upstream/master on your machine; fix it up and push to this branch

cf_xarray/tests/test_accessor.py
kthyng: Ok I added these in https://github.com/xarray-contrib/cf-xarray/pull/234/commits/01da6e3aa7081876415a96b8c43600eea870512a

kthyng:
What's new: https://github.com/xarray-contrib/cf-xarray/pull/234/commits/1835d40640a79a0eb7fa39c4a909b87a676eab5d

kthyng:
Do you want me to update the docs with example on this?

cf_xarray/accessor.py
dcherian: oh so that's a problem for the `standard_name` checks too then

Should we instead have
```suggestion
    my_custom_criteria = {
        "ssh": {
            "standard_name": {"regex": ("sea_surface_elev*",)},
            "name": {"exact": ("sea_surface_elevation",)},  # variable name
        },
    }
```

so there's direct control of regex vs exact matches

kthyng: Yes that is a nice way to make it explicit. Doing this here: https://github.com/xarray-contrib/cf-xarray/pull/234/commits/a6a527764e0fd0fcfd3811c810be42fd21482c7f

kthyng:
I hesitate to say it since in the other PR I was working on I was just doing things wrong with my environments, but I ran the tests locally for this one, in the correct environment I am pretty sure, and they ran ok (except one xfail under test_units.py which must be separate from this work).

kthyng:
I guess the error is:
```
cf_xarray/accessor.py:230: in _get_custom_criteria
    "|".join(expected_vals), obj[var].attrs.get(criterion, "")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = ChainMap(('sea_surface_elev*',))

    def __iter__(self):
        d = {}
        for mapping in reversed(self.maps):
>           d.update(mapping)                   # reuses stored hash values if possible
E           ValueError: dictionary update sequence element #0 has length 17; 2 is required
```
but it doesn't come up in my local testing, and I thought I was using `ChainMap` correctly based on my local work. Is there some nuance I am missing?

dcherian:
It's failing locally here too with python 3.8 . WHat python version are you running. It looks like a tuple was passed to `ChainMap` but it expects mappings like `dict`

kthyng:
I see that the tests pass for Python 3.9 but not for 3.7 or 3.8.

kthyng:
Ok I am not using `ChainMap` anymore and it looks like it is working.

cf_xarray/accessor.py
```diff
@@ -170,6 +180,69 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+        criteria = OPTIONS["custom_criteria"]
+
+    if criteria is not None:
+        criteria = always_iterable(criteria, allowed=(tuple, list, set))
+
+    criteria = ChainMap(*criteria)
+
+    results: Set = set()
+    if key in criteria:
+        for criterion, expected in criteria[key].items():
```
dcherian, line 223:
```suggestion
        for criterion, kind in criteria[key].items():
```

cf_xarray/accessor.py
```diff
@@ -170,6 +180,69 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+            for var in obj.variables:
+                # Treat expected as regex
+                expected_vals = [*expected.values()][0]
```
dcherian, line 226: Can we not just use `kind.values()` instead?

cf_xarray/accessor.py
```diff
@@ -170,6 +180,69 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+                # use regex to match
+                if "regex" in expected:
```
dcherian, line 228:
```suggestion
                if kind == "regex":
```

cf_xarray/accessor.py
```diff
@@ -170,6 +180,69 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+                    if re.match(
+                        "|".join(expected_vals), obj[var].attrs.get(criterion, "")
+                    ):
+                        results.update((var,))
+                    # also check name specifically since not in attributes
+                    elif criterion == "name" and re.match("|".join(expected_vals), var):
+                        results.update((var,))
+                else:  # exact matches only
```
dcherian, line 236:
```suggestion
                elif kind == "exact":  # exact matches only
```

cf_xarray/accessor.py
```diff
@@ -170,6 +180,69 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+                    if obj[var].attrs.get(criterion, "") in expected_vals:
+                        # if re.match("|".join(expected), obj[var].attrs.get(criterion, "")):
+                        results.update((var,))
+                    # also check name specifically since not in attributes
+                    elif criterion == "name" and var in expected_vals:
+                        results.update((var,))
```
dcherian, line 242: Lets raise an error here if kind is not "regex" or "exact" (and also test for that error)

dcherian (commented):
Looks really close, I have a few suggestions

cf_xarray/accessor.py
kthyng: This leads to an error in the following regex line (variable still called "expected" instead of "kind"). 

image

because `expected_vals` is not just a list or tuple:
```
ipdb>  expected_vals
dict_values([('sea_surface_elev*',)])
```

So I have that particular syntax in there to try to work around the dictionary structure since the ChainMap approach didn't work in Python 3.7 or 3.8

kthyng: So I didn't change this other than the variable name change you suggested.

cf_xarray/accessor.py
kthyng: This is in https://github.com/xarray-contrib/cf-xarray/pull/234/commits/e15a2a0adb6367aae0ed122bf79e322b22a6d51c

cf_xarray/accessor.py
```diff
@@ -170,6 +180,76 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+
+    results: Set = set()
+    if key in criteria:
+        for criterion, kind in criteria[key].items():
+            for var in obj.variables:
+                # Treat expected as regex
+                kind_vals = [*kind.values()][0]
+                # use regex to match
+                if "regex" in kind:
+                    if re.match("|".join(kind_vals), obj[var].attrs.get(criterion, "")):
```
dcherian, line 229: > "|".join(kind_vals)

Shall we just expect the user to pass a single string then? Seems like we can expect them to write down the correct rege,x.

cf_xarray/accessor.py
```diff
@@ -170,6 +180,76 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+                        # if re.match("|".join(expected), obj[var].attrs.get(criterion, "")):
+                        results.update((var,))
+                    # also check name specifically since not in attributes
+                    elif criterion == "name" and var in kind_vals:
+                        results.update((var,))
+                else:
+                    raise KeyError(
+                        "User-defined custom_criteria must be defined and must be of the form "
+                        "`{'custom variable name': "
+                        "{'attribute to check': {'regex': ('regex-style pattern*',)}, "
```
dcherian, line 245:
```suggestion
                        "{'attribute to check': {'regex': 'regex-style pattern*'}, "
```

This is what I mean above

cf_xarray/tests/test_accessor.py
```diff
@@ -1237,6 +1237,66 @@ def test_cmip6_attrs():
     assert da.cf.axes["Y"] == ["nlat"]
 
 
+def test_custom_criteria():
+    my_custom_criteria = {
+        "ssh": {
+            "standard_name": {"regex": ("sea_surface_elev*",)},
```
dcherian, line 1243: Sorry to keep flip-flopping on this but another way would be

``` python
ssh = {"standard_name": "sea_surface_elev*|sea_surface_height"}
```

I think this contains both cases pretty nicely. We just expect the users to always provide regex-like input if they want multiple options for "standard name" (i.e. with `|`)

What do you think?

kthyng: Yes now that you say it, this is simpler rather than building in the structure. But, if the user wants an exact match we still need a separate case for that don't we?

kthyng: I'll experiment with this — I suppose the user could have to use "$" at the end of the string to make an exact match with regex.

kthyng: The new version is much simpler — I like it a lot more. It will be easy with enough examples to demo.

cf_xarray/accessor.py
```diff
@@ -170,6 +180,57 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+    if criteria is not None:
+        criteria = always_iterable(criteria, allowed=(tuple, list, set))
+
+    criteria = ChainMap(*criteria)
+
+    results: Set = set()
+    if key in criteria:
+        for criterion, patterns in criteria[key].items():
+            for var in obj.variables:
+                # Treat expected as regex
```
dcherian, line 225:
```suggestion
```

cf_xarray/tests/test_accessor.py
```diff
@@ -1237,6 +1237,68 @@ def test_cmip6_attrs():
...
+    ds["sea_surface_elevation123"] = ("dim", np.arange(10))
+    # Since this will not match, this should error
+    with pytest.raises(KeyError):
+        ds.cf["ssh"]
+
+    # will select only one variable here since exact match
+    ds = xr.Dataset()
+    ds["winds"] = ("dim", np.arange(10), {"standard_name": "wind_speed"})
+    ds["gusts"] = ("dim", np.arange(10), {"standard_name": "wind_speed_of_gust"})
+    ds.cf["wind_speed"]
```
dcherian, line 1282:
```suggestion
    assert_identical(ds.cf["wind_speed"], ds["winds"])
```

cf_xarray/tests/test_accessor.py
```diff
@@ -1237,6 +1237,68 @@ def test_cmip6_attrs():
...
+
+    # Match by standard_name regex match
+    ds = xr.Dataset()
+    ds["elev"] = ("dim", np.arange(10), {"standard_name": "sea_surface_elevBLAH"})
+    xr.testing.assert_identical(ds.cf["ssh"], ds["elev"])
+
+    # Match by standard_name exact match
+    ds = xr.Dataset()
+    ds["salinity"] = ("dim", np.arange(10), {"standard_name": "salinity"})
+    xr.testing.assert_identical(ds.cf["salt"], ds["salinity"])
```
dcherian, line 1269:
```suggestion
    assert_identical(ds.cf["salt"], ds["salinity"])
```

cf_xarray/tests/test_accessor.py
```diff
@@ -1237,6 +1237,68 @@ def test_cmip6_attrs():
...
+    ds["sea_surface_height"] = (
+        "dim",
+        np.arange(10),
+        {"standard_name": "sea_surface_elevBLAH"},
+    )
+    # Since there are two variables, this should error
+    with pytest.raises(KeyError):
+        ds.cf["ssh"]
+    # But the following should work instead given the two ssh variables
+    assert isinstance(ds.cf[["ssh"]], Dataset)
```
dcherian, line 1296:
```suggestion
```

cf_xarray/tests/test_accessor.py
```diff
@@ -1237,6 +1237,68 @@ def test_cmip6_attrs():
...
+    xr.testing.assert_identical(
```
dcherian, line 1297:
```suggestion
    assert_identical(
```

cf_xarray/accessor.py
```diff
@@ -170,6 +180,56 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+) -> List[str]:
+    """
+    Translate from axis, coord, or custom name to variable name optionally
+    using ``custom_criteria``
+
+    Parameters
+    ----------
+    obj : DataArray, Dataset
+        DataArray belonging to the coordinate to be checked
+    key : str, ["X", "Y", "Z", "T", "longitude", "latitude", "vertical", "time"], or a key from user-defined custom_criteria
```
dcherian, line 194:
```suggestion
    key : str
```

cf_xarray/accessor.py
```diff
@@ -170,6 +180,56 @@ def _get_groupby_time_accessor(var: Union[DataArray, Dataset], key: str) -> List
...
+    obj: Union[DataArray, Dataset], key: str, criteria=None
+) -> List[str]:
+    """
+    Translate from axis, coord, or custom name to variable name optionally
+    using ``custom_criteria``
+
+    Parameters
+    ----------
+    obj : DataArray, Dataset
+        DataArray belonging to the coordinate to be checked
```
dcherian, line 193:
```suggestion
```

dcherian:
Nice! I think this is ready to go in.

@malmans2 can you take a quick look please?

cf_xarray/tests/test_accessor.py
```diff
@@ -1237,6 +1237,67 @@ def test_cmip6_attrs():
...
+        "wind_speed": {
+            "standard_name": "wind_speed$",
+        },
+    }
+    cf_xarray.accessor.set_options(my_custom_criteria)
+
+    # Match by name regex match
+    ds = xr.Dataset()
+    ds["salinity"] = ("dim", np.arange(10))
+    xr.testing.assert_identical(ds.cf["salt"], ds["salinity"])
```
dcherian, line 1259:
```suggestion
    assert_identical(ds.cf["salt"], ds["salinity"])
```

cf_xarray/tests/test_accessor.py
```diff
@@ -1237,6 +1237,67 @@ def test_cmip6_attrs():
...
+
+    # Match by standard_name regex match
+    ds = xr.Dataset()
+    ds["elev"] = ("dim", np.arange(10), {"standard_name": "sea_surface_elevBLAH"})
+    xr.testing.assert_identical(ds.cf["ssh"], ds["elev"])
```
dcherian, line 1264:
```suggestion
    assert_identical(ds.cf["ssh"], ds["elev"])
```

cf_xarray/accessor.py
```diff
@@ -65,6 +65,16 @@
...
 ATTRS["vertical"] = ATTRS["Z"]
 
+OPTIONS: MutableMapping[str, Any] = {"custom_criteria": []}
+
+
+def set_options(custom_criteria):
+    OPTIONS["custom_criteria"] = (
+        always_iterable(custom_criteria, allowed=(tuple, list, set))
+        + OPTIONS["custom_criteria"]
+    )
```
malmans2, line 75: I think it would fail with if custom criteria is a tuple or a set.
```suggestion
def set_options(custom_criteria):
    OPTIONS["custom_criteria"].extend(
        always_iterable(custom_criteria, allowed=(tuple, list, set))
    )
```

cf_xarray/accessor.py
```diff
@@ -586,6 +650,8 @@ def check_results(names, key):
         measures = []
         warnings.warn("Ignoring bad cell_measures attribute.", UserWarning)
 
+    custom_criteria = ChainMap(*OPTIONS["custom_criteria"])
```
malmans2, line 653: `ChainMap(*OPTIONS["custom_criteria"])` is always called both here and by `_get_custom_criteria`.
I'm wondering whether it makes sense to try to optimize it .
Maybe we could do something like what we are doing with `_all_cell_measures` and `_get_all_cell_measures`, where we define `_all_cell_measures = None` when we initialize the accessor, and we call `_get_all_cell_measures` only `if _all_cell_measures is None`)

cf_xarray/tests/test_accessor.py
```diff
@@ -1237,6 +1237,67 @@ def test_cmip6_attrs():
...
+        },
+        "salt": {
+            "standard_name": "salinity",
+            "name": "sal*",
+        },
+        "wind_speed": {
+            "standard_name": "wind_speed$",
+        },
+    }
+    cf_xarray.accessor.set_options(my_custom_criteria)
```
malmans2, line 1254: `my_custom_criteria` is set for all the pytest functions called after this, right? Not sure if we want this... 

Also, maybe add a test for `my_custom_criteria` list and/or tuple and/or set as well?

malmans2 (commented):
Looks great!

Should we add a docstring to `set_options`, import it in `__init__`, and add it in the Top-level API?
It would be nice to implement the context manager as well, but that could be a separate PR?

Final thought, I think right now we can add custom criteria, but we can't remove/drop anything (unless we reset everything re-importing cf-xarray, right?). Should we try to address it?

cf_xarray/accessor.py
dcherian, line 75:
```suggestion
    OPTIONS["custom_criteria"] = always_iterable(custom_criteria, allowed=(tuple, list, set))
```

How about we do this for now, so we can always reset it?

cf_xarray/accessor.py
dcherian: I think this is OK for now; I want to refactor this method eventually.

dcherian:
> It would be nice to implement the context manager as well, but that could be a separate PR?

Yes see #237 

> It would be nice to implement the context manager as well, but that could be a separate PR?

Oh hmm..

cf_xarray/accessor.py
malmans2: Sounds good!

cf_xarray/tests/test_accessor.py
kthyng: I don't see how `my_custom_criteria` could be anything besides a dictionary and still have the necessary structure containing the custom name, attribute to search, and patterns to match with. Can we just limit it to being a dict?

malmans2: I think the way is set up right now, `my_custom_criteria` could be a dictionary or an iterable of dictionaries.
For example, you could pass multiple dictionaries at once. So, using `my_custom_criteria` defined in `test_accessor`, this should work too: `cf_xarray.accessor.set_options([my_custom_criteria])`

I guess limit to `dict` is OK, but at that point I think I'm missing why we are using `ChainMap` under the hood and I guess `OPTIONS["custom_criteria"]` should be just a dict as well?

dcherian: I think we want to support a iterable of dictionaries too.

kthyng: Oh I see, I was trying to imagine how to have a list of the names and patterns, etc, not a super list of dictionaries. I am new to ChainMap and have just been trying to follow what is already in the code for that.

I can add a test for a list and tuple of dictionaries.

kthyng: I just added two tests, one for a list and one for a tuple, but the tuple one doesn't work. If you can see an obvious problem please let me know, otherwise I'll try to figure it out in a few hours when I'm back to this work.

malmans2: I think is this: https://github.com/xarray-contrib/cf-xarray/pull/234#discussion_r656965548

malmans2: I.e., you can't do list + tuple (but you can use list.extend(tuple)).

We decided to go with this implementation though: https://github.com/xarray-contrib/cf-xarray/pull/234#discussion_r657052234

kthyng: Ah, thanks, I see. Ok I will commit @dcherian's implementation and see if the tests run.

kthyng:
It looks to me like everything is working and the comments have been accounted for one way or another. What do you think @dcherian and @malmans2?

dcherian:
Thanks @kthyng this is a great PR!

kthyng:
woohoo! This is going to be a big help in several packages I'm working on now. Thanks for your in-depth help @dcherian and checks at the end @malmans2!

PR #238, opened 2021-06-16
Added another salinity unit

kthyng:
I added another preprocessor (in addition to modifying %) to replace the string "1e-3" with "psu". Also added a test.

I was not able to figure out another way to add the unit because it always interpreted the numbers as a scaling factor and couldn't figure out how to deal with that.

Note that I could instead change 1e-3 to 1, which is more cf-compliant as I understand it.

### DIFF
--- a/cf_xarray/units.py
+++ b/cf_xarray/units.py
@@ -41,7 +41,7 @@ units.define(
     "degrees_east = degree = degrees_E = degreesE = degree_east = degree_E = degreeE"
 )
 units.define("@alias meter = gpm")
-units.define("practical_salinity_unit = [] = psu")
+units.define("practical_salinity_unit = [] = psu  = 1 PSS-78 = 1 Practical Salinity Scale 78 = 1.e-3 = 1E-3")
 
 # Enable pint's built-in matplotlib support
 try:
--- a/cf_xarray/units.py
+++ b/cf_xarray/units.py
@@ -15,9 +15,11 @@ from pint import (  # noqa: F401
 )
 
 # Create registry, with preprocessors for UDUNITS-style powers (m2 s-2) and percent signs
+# also change salinity unit 1e-3 to psu
 units = pint.UnitRegistry(
     autoconvert_offset_to_baseunit=True,
     preprocessors=[
+        lambda string: string.replace("1e-3", "psu"),
         functools.partial(
             re.compile(
                 r"(?<=[A-Za-z])(?![A-Za-z])(?<![0-9\-][eE])(?<![0-9\-])(?=[0-9\-])"
@@ -41,7 +43,7 @@ units.define(
     "degrees_east = degree = degrees_E = degreesE = degree_east = degree_E = degreeE"
 )
 units.define("@alias meter = gpm")
-units.define("practical_salinity_unit = [] = psu  = 1 PSS-78 = 1 Practical Salinity Scale 78 = 1.e-3 = 1E-3")
+units.define("practical_salinity_unit = [] = psu")
 
 # Enable pint's built-in matplotlib support
 try:
--- a/cf_xarray/units.py
+++ b/cf_xarray/units.py
@@ -19,7 +19,7 @@ from pint import (  # noqa: F401
 units = pint.UnitRegistry(
     autoconvert_offset_to_baseunit=True,
     preprocessors=[
-        lambda string: string.replace("1e-3", "psu"),
+        lambda string: string.replace("1e-3", "1"),
         functools.partial(
             re.compile(
                 r"(?<=[A-Za-z])(?![A-Za-z])(?<![0-9\-][eE])(?<![0-9\-])(?=[0-9\-])"
--- a/cf_xarray/tests/test_units.py
+++ b/cf_xarray/tests/test_units.py
@@ -44,6 +44,12 @@ def test_psu_unit():
     assert str(x.units) == "practical_salinity_unit"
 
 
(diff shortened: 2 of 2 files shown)


cf_xarray/units.py
```diff
@@ -15,9 +15,11 @@
 )
 
 # Create registry, with preprocessors for UDUNITS-style powers (m2 s-2) and percent signs
+# also change salinity unit 1e-3 to psu
 units = pint.UnitRegistry(
     autoconvert_offset_to_baseunit=True,
     preprocessors=[
+        lambda string: string.replace("1e-3", "psu"),
```
dcherian, line 22: i guess this is the best approach. unless @jthielen has ideas.

kthyng: Do you like psu vs. changing to "1" as the unit?

Also I am able to use the same registry with `pint-pandas` so that is really handy! And import orders matters completely. I had inadvertently done the import order wrong because I had dependencies within something I imported first that messed up the order.

dcherian: I think I prefer `"1"`.

> Also I am able to use the same registry with pint-pandas so that is really handy! 

nice!

> And import orders matters completely. 

Looks like pint-xarray should add a little FAQ entry on checking this. cc @keewis

kthyng: Ok salinity units of 1e-3 are now changed to 1.

I'm happy with this when others are!

kthyng:
Hm, the tests run locally for me. Not sure why they aren't working in the CI.

dcherian:
It's testing with pint 0.17; is that the version you have?

cf_xarray/units.py
keewis: > Looks like pint-xarray should add a little FAQ entry on checking this.

will do. I'm still hoping to eventually figure out a way to completely avoid this, though.

kthyng:
> It's testing with pint 0.17; is that the version you have?

Oops, I didn't have pint installed at all in this environment! Have been working on the code in another env. Didn't realize the tests are skipped if pint isn't installed. 

Maybe converting 1e-3 to psu is better because then it has a real "unit" that can be queried with 
```
    x = units("1e-3")
    x.units
```
whereas converting it to "1" leads to just an integer "1" for x in the above example and `x.units` is not valid. I think I could make it so that "1" is a valid unit then but that would pick up other dimensionless units wouldn't it?

cf_xarray/units.py
dcherian: that would be better!

dcherian:
Ugh OK psu it is...

dcherian:
actually wait, should we just change the test and `assert x == 1`?

kthyng:
> actually wait, should we just change the test and `assert x == 1`?

I'm not sure — I don't have much experience in how `pint` is used generally since I am just starting up with it, but I would think we want the result to have units? I'm not sure if that ends up mattering or not. Or if this would make salinity with these units look like something else that isn't salinity? But maybe that is ok.

kthyng:
Maybe it would be more concerning if salinity was used in calculations or converted to other units but that doesn't really happen in reality right?

dcherian:
Nah. `psu` is just `1` really, so it doesn't matter. Arguably it's more correct to drop the units

dcherian:
What happens if you quantify something with `units="psu"` and another with `units="1"` and tried to subtract them?

keewis:
it's probably best to have "salinity" behave like angles: they're dimensionless so you can add / subtract anything that doesn't have a unit, but you can convert between radians, degree and gon, and adding radians to degree automatically converts one of the quantities (falling back to radians if one of them is not an angle).

That would require a dimensionless "salinity unit" (not sure if that's the best name) and allow defining everything else relative to that:
```
salinity_unit = []
psu = practical_salinity_unit = 1e-3 salinity_unit
```

jthielen:
I concur with @keewis as to the likely best way to represent this (at least as far as I understand the unit...I'm an atmospheric rather than oceanic scientist so it's outside my expertise). 

I'm worried doing this as a universal preprocessing step (like this PR currently has) will break expected behavior by changing something ("1e-3") into something that it isn't usually equivalent to in other contexts. For example, water vapor mixing ratio is commonly in units of "grams/kilogram", which can also be validly encoded as "1e-3". Changing that to "1" would thereby make such quantities incorrect by a factor of 1000.

kthyng:
Thanks so much for your input @keewis and @jthielen! I was also worried about the conversion 1e-3 but wasn't sure if anything else really used it, and couldn't figure out another way to do it. I will try this now.

kthyng:
@keewis Is there a way to make it so "1e-3" can be interpreted as being salinity without messing up other variables that also use 1e-3? I'm guessing not. One of my use cases is to be able to interpret the units "1e-3" from DataFrames or Datasets as salinity units, or at least just units without throwing an errors. See snapshot for visual:
image

I read I could define units locally — would that be better if this will overlap too much with other variables?

dcherian:
Thanks @jthielen!

@kthyng Maybe the best way is to preprocess it yourself
``` python
for name in itertools.chain(
	ds.cf.standard_names["sea_water_salinity"], 
	ds.cf.standard_names["sea_water_practical_salinity"]
):
	ds[name].attrs["units"] = "1"
```

kthyng:
Yeah, I am thinking this isn't going to work out after all. The 1e-3 unit is very confusing. I think I will preprocess that data I'm working with to avoid the problem.

Thanks everyone for your help and time!

PR #239, opened 2021-06-25
added PSU to psu units

kthyng:
This is a baby PR that adds "PSU" to the list of recognized salinity units.


dcherian:
Thanks you should go ahead and merge when tests pass

Issue #122, opened 2021-01-07
Add support for discrete axis, instance dimensions, and index variables?


malmans2:
I ran into this while trying to set up a dataset with a collection of vertical profiles (i.e., a transect).

I think we should consider adding a new axis named "discrete": http://cfconventions.org/cf-conventions/cf-conventions.html#discrete-axis
The dimensions of the discrete axis would then be defined by an attribute named "instance_dimension": http://cfconventions.org/cf-conventions/cf-conventions.html#collections-instances-elements
The "instance_dimension" attribute is assigned to all "index_variables", which are 1D coordinates.

This is an example where I've extracted a transect from a C-grid model. After extraction I've removed the "axis" attributes from all X and Y variables, and I've added the attribute `da.attrs["instance_dimension"] = "station"` to all 1D variables with dimension "station".

![discrete](https://user-images.githubusercontent.com/22245117/103901265-d15f7800-50f0-11eb-83dd-c0475b905d1f.png)

In this scenario, I think the axes should be "Z", "T", and "discrete", where `ds.axes["discrete"] = ["station"]`. Then we should probably also add `ds.cf.index_variables`, which returns all index variables (e.g., lon, lat, label, indexes on the original grid, ....).

There is a global attribute named "featureType":  http://cfconventions.org/cf-conventions/cf-conventions.html#_features_and_feature_types
Not sure whether it would be preferable to add `ds.cf["discrete"]` and `ds.cf.index_variables` only if the attribute is present, and maybe axes ["X", "Y"] and "discrete" should be mutually exclusive?

dcherian:
Yes we should support this "discrete sampling geometry" stuff. 

I am confused about how these are represented however. The CF CDL examples don't use the `discrete` attribute; http://cfconventions.org/cf-conventions/cf-conventions.html#_indexed_ragged_array_representation_of_trajectories but `instance_dimension` is used.

@ocefpaf Can you point us to a "nice" dataset that uses these attributes?

ocefpaf:
> @ocefpaf Can you point us to a "nice" dataset that uses these attributes?

I believe we have some "gold standards" somewhere. Let me check and get back to you.

rsignell-usgs:
[...]

PR #240, opened 2021-06-28
Support indexing by cf_role for DSG datasets

dcherian:
xref #122

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -25,7 +25,7 @@ import xarray as xr
 from xarray import DataArray, Dataset
 from xarray.core.arithmetic import SupportsArithmetic
 
-from .criteria import coordinate_criteria, regex
+from .criteria import cf_role_criteria, coordinate_criteria, regex
 from .helpers import bounds_to_vertices
 from .utils import (
     _get_version,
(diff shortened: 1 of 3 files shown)


cf_xarray/accessor.py
```diff
@@ -372,6 +372,7 @@ def _get_all(obj: Union[DataArray, Dataset], key: str) -> List[str]:
     """
     all_mappers = (
         _get_custom_criteria,
+        functools.partial(_get_custom_criteria, criteria=cf_role_criteria),
```
dcherian, line 375: @kthyng This was really easy!

cf_xarray/accessor.py
```diff
@@ -665,7 +666,7 @@ def check_results(names, key):
             successful[k] = bool(measure)
             if measure:
                 varnames.extend(measure)
-        elif k in custom_criteria:
+        elif k in custom_criteria or k in cf_role_criteria:
```
dcherian, line 669: we should really refactor this.

cf_xarray/accessor.py
kthyng: @dcherian Nice! Not sure what this is exactly, but glad the changes are working for you.

PR #243, opened 2021-06-29
Copy over xarray's set_options context manager

kthyng:
This makes the support for the custom_criteria more robust.

### DIFF
--- a/cf_xarray/__init__.py
+++ b/cf_xarray/__init__.py
@@ -1,6 +1,5 @@
 from .accessor import CFAccessor  # noqa
 from .helpers import bounds_to_vertices, vertices_to_bounds  # noqa
 from .utils import _get_version
-from .options import set_options
 
 __version__ = _get_version()
--- a/cf_xarray/__init__.py
+++ b/cf_xarray/__init__.py
@@ -1,6 +1,6 @@
 from .accessor import CFAccessor  # noqa
 from .helpers import bounds_to_vertices, vertices_to_bounds  # noqa
-from .options import set_options
+from .options import set_options  # noqa
 from .utils import _get_version
 
 __version__ = _get_version()
--- a/cf_xarray/__init__.py
+++ b/cf_xarray/__init__.py
@@ -1,5 +1,6 @@
 from .accessor import CFAccessor  # noqa
 from .helpers import bounds_to_vertices, vertices_to_bounds  # noqa
+from .options import set_options
 from .utils import _get_version
 
 __version__ = _get_version()
--- a/cf_xarray/__init__.py
+++ b/cf_xarray/__init__.py
@@ -1,5 +1,6 @@
 from .accessor import CFAccessor  # noqa
 from .helpers import bounds_to_vertices, vertices_to_bounds  # noqa
 from .utils import _get_version
+from .options import set_options
 
 __version__ = _get_version()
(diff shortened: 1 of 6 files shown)


kthyng:
@dcherian A precommit check doesn't like the import I'm doing for the options but it is necessary for the tests to work. I'm not sure how to get around this.

cf_xarray/options.py
```diff
@@ -0,0 +1,51 @@
...
+        *empty*
+    Or to set global options:
+    >>> xr.set_options(display_width=80)  # doctest: +ELLIPSIS
+    <xarray.core.options.set_options object at 0x...>
+    """
+
+    def __init__(self, **kwargs):
+        self.old = {}
+        for k, v in kwargs.items():
+            if k not in OPTIONS:
```
dcherian, line 37: we should add a test for this.

cf_xarray/options.py
```diff
@@ -0,0 +1,51 @@
...
+                raise ValueError(
+                    f"argument name {k!r} is not in the set of valid options {set(OPTIONS)!r}"
+                )
+            self.old[k] = OPTIONS[k]
+        self._apply_update(kwargs)
+
+    def _apply_update(self, options_dict):
+        OPTIONS.update(options_dict)
```
dcherian, line 45: I think we want the `always_iterable` bit that was deleted earlier. We can add it after a `if key == "custom_criteria"`

cf_xarray/__init__.py
```diff
@@ -1,5 +1,6 @@
 from .accessor import CFAccessor  # noqa
 from .helpers import bounds_to_vertices, vertices_to_bounds  # noqa
+from .options import set_options
```
dcherian, line 3:
```suggestion
from .options import set_options  # noqa
```

this will silence the flake8 warning

cf_xarray/options.py
kthyng: Oops, ok.

cf_xarray/options.py
```diff
@@ -42,6 +39,11 @@ def __init__(self, **kwargs):
         self._apply_update(kwargs)
 
     def _apply_update(self, options_dict):
+        for k, v in options_dict.items():
+            if k == "custom_criteria":
+                OPTIONS["custom_criteria"] = always_iterable(
+                    OPTIONS["custom_criteria"], allowed=(tuple, list)
```
dcherian, line 45:
```suggestion
                options_dict["custom_criteria"] = always_iterable(
                    options_dict["custom_criteria"], allowed=(tuple, list)
```

cf_xarray/options.py
```diff
@@ -42,6 +39,11 @@ def __init__(self, **kwargs):
...
+                )
         OPTIONS.update(options_dict)
```
dcherian, line 47: this will overwrite the earlier value

kthyng: Ok hopefully I fixed that.

dcherian: hmm.. now we're mutating user input. Let's do a `copy.deepcopy(options_dict)` first.

kthyng: https://github.com/xarray-contrib/cf-xarray/pull/243/commits/092d427794947ca021425b4272a49200a966841a ?

dcherian:
Thanks!

Issue #163, opened 2021-02-04
Compute 1d bounds on 4D array


jbusecke:
Thank you for developing this amazing package, I am only slowly appreciating its full amazingness.

I am currently trying to convert an array of layer depths that varies in time and space, which represent cell vertices in the vertical into a `bound` style array. 

This is my original array `lev_vertices`:
![image](https://user-images.githubusercontent.com/14314623/106915887-9131ec00-66d4-11eb-9a03-017b8fc16d2d.png)

For a minimal example you can create a similar array with:
```python
lev_vertices = xr.DataArray(np.arange(9), dims='test').expand_dims(x=2,y=2, time=2)
lev_vertices
```

I would like to convert the dimension `test` into bounds.
This works as expected when operating on a single profile:
```python
cf_xarray.vertices_to_bounds(
    lev_bounds.isel(time=0,x=0, y=0).load(), out_dims=["bnds", "test"]
)
```
![image](https://user-images.githubusercontent.com/14314623/106916549-4d8bb200-66d5-11eb-961a-3fbd4148998c.png)

But I cant figure out how to make this work on the full array...

```python
cf_xarray.vertices_to_bounds(
    lev_vertices.load(), out_dims=["bnds", "test"]
)
```
gives 

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-76-4487305bbcdc> in <module>
----> 1 cf_xarray.vertices_to_bounds(
      2     lev_vertices.load(), out_dims=["bnds", "test"]
      3 )

/scratch/gpfs/jbusecke/conda_tigercpu/envs/aguadv_omz_busecke_2021/lib/python3.8/site-packages/cf_xarray/helpers.py in vertices_to_bounds(vertices, out_dims)
    116         )
    117     else:
--> 118         raise ValueError(
    119             f"vertices format not understood. Got {vertices.dims} with shape {vertices.shape}."
    120         )

ValueError: vertices format not understood. Got ('x', 'y', 'time', 'test') with shape (2, 2, 2, 9).
```

I am wondering if I am missing an option to ignore dimensions here? I was thinking of using xr.map_blocks, but that would impose restrictions on my chunking (which might affect performance). So I first wanted to check in here, and see if there is an easy fix.

jbusecke:
This little snippet actually does exactly what I want: 
```python
[...]

PR #250, opened 2021-07-13
Convert nD bounds arrays to (n-1)D vertex arrays

dcherian:
xref #163 which is the inverse of this


aulemahal (approved):
Nice generalization! Thanks!

(Sorry for the delay, I was on vacation for the last 3 weeks)

dcherian:
thanks for the review!

PR #253, opened 2021-08-03
Add attribute tracking functionality

dcherian:
Simple test of https://github.com/pydata/xarray/pull/5668 (cc @keewis)

This just adds the `cell_methods` attribute for reductions and a placeholder for tracking history.

``` python
import cf_xarray as cfxr
import xarray as xr

da = xr.DataArray([1, 2, 3], dims="time")
with xr.set_options(keep_attrs=cfxr.track_cf_attributes(cell_methods=True)):
    print(da.mean("time").attrs)
```

```
{'cell_methods': 'dim_name: mean ', 'history': None}
```

There is one top-level function
```
cfxr.tracking.track_cf_attributes(cell_methods=True, history=True)
```
This returns a partial function that can be provided to `keep_attrs` i.e. it expects the args `attrs, context`. I think we could add things like `provenance=True` etc. in the future. 

Right now only the private `_tracker` function satisfies the contract expected by `keep_attrs`. Alternately, we could provide `track_cell_methods` `track_history` etc. that satisfy the contract so that users can instead do for fine control
```
xr.set_options(keep_attrs=cfxr.tracking.track_history)
```

I'm interested in opinions on whether we should do this second approach instead.

cc @huard @aulemahal

### DIFF
--- a/cf_xarray/__init__.py
+++ b/cf_xarray/__init__.py
@@ -1,6 +1,8 @@
+from . import tracking  # noqa
 from .accessor import CFAccessor  # noqa
 from .helpers import bounds_to_vertices, vertices_to_bounds  # noqa
 from .options import set_options  # noqa
+from .tracking import track_cf_attributes  # noqa
 from .utils import _get_version
 
 __version__ = _get_version()
--- a/cf_xarray/tracking.py
+++ b/cf_xarray/tracking.py
@@ -4,6 +4,7 @@
 
 import copy
 import functools
+from datetime import datetime
 
 CELL_METHODS = {
     "sum": "sum",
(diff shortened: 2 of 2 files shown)


aulemahal:
While I like the idea of `xr.set_options(keep_attrs=cfxr.tracking.track_history)`, it becomes complex when you want to track both! And in the hypothetical case where we add even more things to track, it'll be a nightmare. So I vote for the single function option!

Otherwise, this functionality is very interesting!

dcherian:
> Right now only the private _tracker function satisfies the contract expected by keep_attrs. 

Actually this is wrong! Both `add_history` and ` add_cell_methods` satisfy the contract, so this was a useless question. I apologize for the noise.

huard (approved):

PR #255, opened 2021-08-09
Fix bounds_to_vertices

aulemahal:
The recent changes to `bounds_to_vertices` (#250) broke the case where `order=None` and `"counterclockwise"` is the good solution. (The tests were testing the case where `order=None` and solution is `"clockwise"`) We realized it in xESMF : pangeo-data/xESMF#112.

I missed that during my review, but the problem is simply a change in the axes order within the function... This adds a test and fixes the bug.

### DIFF
--- a/cf_xarray/tests/test_helpers.py
+++ b/cf_xarray/tests/test_helpers.py
@@ -18,8 +18,14 @@ def test_bounds_to_vertices():
     assert_array_equal(ds.lat.values + 1.25, lat_c.values[:-1])
 
     # 2D case, CF- order
-    lat_c = cfxr.bounds_to_vertices(mollwds.lat_bounds, bounds_dim="bounds")
-    assert_equal(mollwds.lat_vertices, lat_c)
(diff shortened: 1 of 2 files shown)


cf_xarray/tests/test_helpers.py
```diff
@@ -18,8 +18,14 @@ def test_bounds_to_vertices():
     assert_array_equal(ds.lat.values + 1.25, lat_c.values[:-1])
 
     # 2D case, CF- order
-    lat_c = cfxr.bounds_to_vertices(mollwds.lat_bounds, bounds_dim="bounds")
-    assert_equal(mollwds.lat_vertices, lat_c)
+    lat_ccw = cfxr.bounds_to_vertices(
```
dcherian, line 21: should we keep the clockwise test case too?

aulemahal: I think so, so now both cases are tested. Also because the new test only tests that cf_xarray is coherent with itself. The second one further down tests against verified bounds and vertices.

PR #267, opened 2021-11-24
Drop missing variables

### LINKED ISSUE #254: Remove nonexistent variable names from properties like bounds, cell_measures, formula_terms
I am using `cf-xarray` to write a custom accessor related to coordinate bounds.

The Problem:
* I found that if you drop existing bounds vars, the key-value pair for is still retained in `ds.cf.bounds` dict. This test was performed on a real-world `.nc` file.

Expected Behavior:
* The corresponding key-value pair for a dropped bound var should be dropped from the `ds.cf.bounds` dict.

Steps to Reproduce:
```
import cf_xarray as cfxr
import xarray as xr

$ ds = xr.open_dataset(...)
$ ds.data_vars
Data variables:
    time_bnds  (time, bnds) datetime64[ns] 1850-01-01 1850-02-01 ... 2006-01-01
    lat_bnds   (lat, bnds) float64 -90.0 -89.38 -89.38 ... 89.38 89.38 90.0
    lon_bnds   (lon, bnds) float64 -0.9375 0.9375 0.9375 ... 357.2 357.2 359.1
    ts         (time, lat, lon) float32 252.7 252.7 252.7 ... 248.2 248.2 248.2

$ ds.cf.bounds
{'longitude': ['lon_bnds'],
 'X': ['lon_bnds'],
 'Y': ['lat_bnds'],
 'latitude': ['lat_bnds'],
 'lon': ['lon_bnds'],
 'lat': ['lat_bnds'],
 'time': ['time_bnds'],
 'T': ['time_bnds']}

$ ds = ds.drop_vars("lon_bnds")
$ ds.data_vars
Data variables:
    time_bnds  (time, bnds) datetime64[ns] 1850-01-01 1850-02-01 ... 2006-01-01
    lat_bnds   (lat, bnds) float64 -90.0 -89.38 -89.38 ... 89.38 89.38 90.0
    ts         (time, lat, lon) float32 252.7 252.7 252.7 ... 248.2 248.2 248.2

# "longitude" : ["lon_bnds"] should be dropped from the dict
$ ds.cf.bounds
{'longitude': ['lon_bnds'],
 'X': ['lon_bnds'],
 'Y': ['lat_bnds'],
 'latitude': ['lat_bnds'],
 'lon': ['lon_bnds'],
 'lat': ['lat_bnds'],
 'time': ['time_bnds'],
 'T': ['time_bnds']}

```

https://github.com/xarray-contrib/cf-xarray/blob/16d8c399048ba168d65eaabe1d3402922b1145c7/cf_xarray/accessor.py#L1865-L1883

Alternatives I've Considered:
* I loop through a statically defined list of coordinates and use `ds.cf.bounds.get_bounds()` instead.
```
    missing_bounds = []
    for coord in coords:
        try:
            dataset.cf.get_bounds(coord)
        except KeyError:
            missing_bounds.append(coord)
```

malmans2:
Closes #254 

I haven't tried with other attributes, but it looks like introducing a decorator that only returns existing keys would work well.
Thoughts? If you think that this is the way to go, I'll try to implement it for `cell_measures`, `formula_terms`, ...

cc: @dcherian @tomvothecoder @huard

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -413,6 +413,14 @@ def _variables(func: F) -> F:
     return cast(F, wrapper)
 
 
+def _must_exist(func: F) -> F:
+    @functools.wraps(func)
+    def wrapper(obj: Union[DataArray, Dataset], key: str) -> List[DataArray]:
+        return [k for k in func(obj, key) if k in obj]
+
+    return cast(F, wrapper)
+
+
 def _single(func: F) -> F:
     @functools.wraps(func)
     def wrapper(obj: Union[DataArray, Dataset], key: str):
@@ -1896,7 +1904,8 @@ class CFDatasetAccessor(CFAccessor):
         keys = self.keys() | set(obj.variables)
 
         vardict = {
-            key: apply_mapper(_get_bounds, obj, key, error=False) for key in keys
+            key: apply_mapper(_must_exist(_get_bounds), obj, key, error=False)
+            for key in keys
         }
 
         return {k: sorted(v) for k, v in vardict.items() if v}
@@ -1915,7 +1924,9 @@ class CFDatasetAccessor(CFAccessor):
         DataArray
         """
 
-        return apply_mapper(_variables(_single(_get_bounds)), self._obj, key)[0]
+        return apply_mapper(
+            _variables(_single(_must_exist(_get_bounds))), self._obj, key
+        )[0]
 
     def get_bounds_dim_name(self, key: str) -> str:
         """
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -413,20 +413,16 @@ def _variables(func: F) -> F:
     return cast(F, wrapper)
 
 
-def _raise_if_not_single(key, results):
-    if len(results) > 1:
-        raise KeyError(
-            f"Multiple results for {key!r} found: {results!r}. I expected only one."
-        )
-    elif len(results) == 0:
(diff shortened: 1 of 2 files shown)


cf_xarray/accessor.py
```diff
@@ -1924,9 +1926,10 @@ def get_bounds(self, key: str) -> DataArray:
         DataArray
         """
 
-        return apply_mapper(
-            _variables(_single(_must_exist(_get_bounds))), self._obj, key
-        )[0]
+        results = self.bounds.get(key, [])
+        _raise_if_not_single(key, results)
```
dcherian, line 1930: Not sure if this is right. For a staggered grid you could have multiple lat/lon with different bounds though I don't know that anyone ever does that :)

malmans2: Mmmm, not sure. I think we would have to change how we pare `da.attrs['bounds']`, because I think right now we expect a single string.

malmans2: Ohh actually you're right, for example key="longitude" could be associated to multiple bounds. The problem is that I think other functions use this and need a single DataArray. Maybe it's better to address this in a separate PR...

dcherian: Lets open an issue for that. And here lets not raise_if_single...

huard (approved):

malmans2:
This should be good to go!

dcherian:
Thanks Mattia

PR #269, opened 2021-12-09
add xarray as a dependency

ocefpaf:
@dcherian is this OK? Should we add a min version?


dcherian:
No min version necessary AFAIK. Thanks! This was an oversight.

PR #271, opened 2021-12-13
Allow hashable keys (not only strings) 

### LINKED ISSUE #229: Support DataArrays with names that are not string
Right now pretty much everything is broken if any `DataArray` in a `Dataset` has a name that is not a string.
However, I don't think we have to introduce too many changes to fix this, it involves something very similar to what we did in #227.

What do you think, does it makes sense to support these kind of `Datasets`?

aulemahal:
Fixes #229.

I modified `apply_mapper` to fit with xarray's criterion for a `DataArray` name, meaning that `key` must be hashable (and not only simply a string). I added a test for the `_to_temp_dataset()` case. All other tests pass.

As far as I understand, the `key` is tested either against string parsed from the attributes or against variable names in the dataset. In the first case, passing a non-string doesn't raise anything but it's also useless. In the second case, that's precisely the point : "being hashable" is the same constraint as in xarray.

But maybe others have a better idea of the edge cases? I can add more tests, I just don't know what to test!

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -89,10 +89,10 @@ def apply_mapper(
     results for a good key.
     """
 
-    if not isinstance(key, collections.abc.Hashable):
+    if not isinstance(key, Hashable):
         if default is None:
             raise ValueError(
-                "`default` must be provided when `key` is not hashable (not a valid DataArray name)."
+                "`default` must be provided when `key` is not not a valid DataArray name (of hashable type)."
             )
         return list(always_iterable(default))
 
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1,4 +1,3 @@
-import collections.abc
 import functools
 import inspect
 import itertools
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1,3 +1,4 @@
+import collections.abc
 import functools
 import inspect
 import itertools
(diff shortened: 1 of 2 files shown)


cf_xarray/accessor.py
```diff
@@ -88,9 +89,11 @@ def apply_mapper(
     results for a good key.
     """
 
-    if not isinstance(key, str):
+    if not isinstance(key, collections.abc.Hashable):
         if default is None:
-            raise ValueError("`default` must be provided when `key` is not a string.")
+            raise ValueError(
+                "`default` must be provided when `key` is not hashable (not a valid DataArray name)."
```
dcherian, line 95:
```suggestion
                "`default` must be provided when `key` is not not a valid DataArray name (of hashable type)."
```

cf_xarray/accessor.py
```diff
@@ -88,9 +89,11 @@ def apply_mapper(
     results for a good key.
     """
 
-    if not isinstance(key, str):
+    if not isinstance(key, collections.abc.Hashable):
```
dcherian, line 92:
```suggestion
    if not isinstance(key, Hashable):
```

would this just work?

dcherian (commented):
For more tests you could add
`airds.air._to_temp_dataset()` to the `datasets` definition at the top of `test_accessor.py`

aulemahal:
I had to add some conditions in the tests in order to introduce the temp dataset, but it seems to work!

dcherian:
Great thanks @aulemahal

PR #273, opened 2021-12-15
Remove pkg resources

ocefpaf:
@dcherian this is not ready yet but I want to place it here so I don't forget.

Relying on `pkg_resources` can make the initial import of cf_xarray quite slow depending on the packages you have installed. This PR changes the logic a bit to avoid relying on that but rather make use of  setuptools-scm file `_version.py`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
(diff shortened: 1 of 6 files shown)


ocefpaf:
@dcherian this is ready for a review.

dcherian:
Seems  fine  to me! Thanks!

PR #275, opened 2021-12-16
single source for metadata

ocefpaf:
@dcherian the URL in the setup.cfg is for the GH repo but the one in the setup.py is for the docs. Which one do you want to keep?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,6 +2,7 @@
 name = cf_xarray
 license = Apache
 url = https://github.com/xarray-contrib/cf-xarray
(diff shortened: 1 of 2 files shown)


dcherian:
Lets do the  docs. thanks

ocefpaf:
> Lets do the docs. thanks

Done!

keewis:
for reference, you can set multiple urls:
```ini
url = https://github.com/xarray-contrib/cf-xarray
project_urls =
  Documentation=https://cf-xarray.readthedocs.io
```
see [here](https://pypi.org/project/pint-xarray/) for an example on PyPI

PR #278, opened 2022-01-03
custom unit formatter

keewis:
The combination of `pint`'s new-ish custom formatter feature and `xclim`'s unit formatting function (see Ouranosinc/xclim#780). Only the short formatter for now because `pint` does not pass the unit modifier to the formatter (i.e. to support both short and long version, we need a change to `pint` first).

- [x] Tests added
- [x] fully documented
- [x] changelog updated
- [x] licensing: how do you want to handle this? At the moment there's a module docstring declaring everything in the module to be vendored from `metpy`, but that's not true anymore.

cc @aulemahal


aulemahal:
Nice! If your last point was about "sublicensing" xclim, I don't think it's really needed. I mean, as a core dev of xclim, I am totally ok with cf-xarray simply copying the code. To be certain we can ask @huard, who I think wrote this function (`pint2cfunits`).

keewis:
that was directed at @dcherian: I only have experience with vendoring code (which is including the code *without* asking for permission) so I wasn't sure what to do in this case.

A simple comment stating that "this was moved here from xclim", or modifying the notice about `metpy` to not cover everything in the module might probably suffice?

dcherian:
Not sure what the right answer is.

Practically,  I think modifying the notice at the top to specifically mention this function should be OK? We can also add a. comment above this function to make it extra clear.

It does seem like we need to include the text of the BSD-3-clause license somewhere

keewis:
not sure if that's the best way, but I think if the `xclim` developers (in this case @huard?) tell us it's fine without that we would be relicensing it to `cf-xarray`'s license?

dcherian:
xclim and cf_xarray are both apache-2 and apache-2 allows redistributing as long. as the new. license is compatible with apache-2

The super safe thing would be to create a new  `formatting.py` and  import it here.

keewis:
> The super safe thing would be to create a new `formatting.py` and import it here.

or we reorganize the current one to not have the notice at the top, but divide it into sections containing code from the same source. For now, this is easy: there's no dependency between the unit parsing and the unit formatting code.

dcherian:
OK sounds good to me.

keewis:
not sure what the issue with RTD is, but in any case this should be ready for another round of reviews.

dcherian:
LGTM.

Is pint likely to support passing modifiers soon?  Should we instead register `cfshort` and `cflong`?

keewis:
that depends on whether or not I can send in a PR... I think that should not be too much effort, but I won't really know until I work on it

dcherian:
Thanks @keewis. This is a great contribution

huard:
For the record, I'm happy to see this go into xarray.

PR #280, opened 2022-01-08
Upgrade typing syntax

dcherian:


malmans2:
I forgot to as a while ago... Should we add `xarray` [here](https://github.com/xarray-contrib/cf-xarray/blob/44ef3517d659c432efd998c2fc761f045feef0ab/.pre-commit-config.yaml#L25)?
I don't think we are checking xarray objects right now.

dcherian:
nice. thanks!

dcherian:
I'm getting some weird internal error from mypy so skipping for now

Issue #225, opened 2021-05-18
updates to unit support


dcherian:
Current todo list remaining:

- [ ] add documentation
  - Examples of using cf-xarray and pint-xarray together
  - Instructing users to either import cf-xarray before pint-xarray or set 
  - List current known shortfalls of this registry compared to full UDUNITS support (namely case sensitivity)
- [ ] add tests of this registry's quantities being used within xarray, and not just correctly parsing units with the customized features.

_Originally posted by @jthielen in https://github.com/xarray-contrib/cf-xarray/pull/197#issuecomment-816038767_

kthyng:
I've run across this discussion and work now and am interested in using it! Is the pint unit stuff sort of separate from the other cf-xarray stuff? I am looking to interpret units in xarray Datasets but with units including psu and for example m s-1.

keewis:
indeed, the only thing it does for now is provide a special `pint.UnitRegistry` object so datasets with `cfunits` style `"units"` attributes can be used with `pint-xarray`.

The documentation for this feature of `cf-xarray` is still missing, though, and we need to get `pint`'s application registry to work reliably but it should be usable right now. In any case I encourage you to try it out and report any issues you run into.

kthyng:
Thanks @keewis! I haven't actually used `pint` before today so it's slow going, but I'll see what I can figure out.

keewis:
if you're willing to take the time I'd also be interested in feedback on what you found difficult to understand: we didn't check the documentation of `pint-xarray` in that regard, yet.

kthyng:
I've gotten a little bit into this and get something to work, so that has been exciting. I've immediately hit another unit problem though: salinity units as 1e-3. Is this one that can be added into the registry here?

keewis:
I don't know a lot about the units used in oceanography (for example, what is the difference between those and the `"psu"` that are already in the registry?) so I will leave this request to @dcherian or anyone else with more knowledge.

In general, if you can provide a unit name and the dimensionality / scale I think that should be possible.

kthyng:
Here is just a basic example I was able to cobble together from the `pint-xarray` docs and looking at the commits in the new unit changes here in cf-xarray got me far:

```
import pint
# order apparently matters
from cf_xarray.units import units
import pint_xarray

ds = xr.open_dataset([your dataset])
ds = ds.pint.quantify()  # to interpret units
# units like m s-1 will be interpretable by pint with the cf_xarray unit registry
[...]

PR #284, opened 2022-01-12
More unit definitions from xclim

dcherian:
I needed the "degC = celsius"  aliases so I copied that and more from xclim.

cc @jthielen

xref #225

### DIFF
--- a/cf_xarray/units.py
+++ b/cf_xarray/units.py
@@ -97,10 +97,10 @@ units.define("@alias degK = deg_K")
 units.define("@alias day = d")
 units.define("@alias hour = h")  # Not the Planck constant...
 units.define(
-    "degrees_north = 1 * degree = degrees_north = degrees_N = degreesN = degree_north = degree_N = degreeN"
+    "degrees_north = degree = degrees_north = degrees_N = degreesN = degree_north = degree_N = degreeN"
 )
 units.define(
-    "degrees_east = 1 * degree = degrees_east = degrees_E = degreesE = degree_east = degree_E = degreeE"
+    "degrees_east = degree = degrees_east = degrees_E = degreesE = degree_east = degree_E = degreeE"
 )
 units.define("[speed] = [length] / [time]")
 # ----- end block copied from xclim
--- a/cf_xarray/units.py
+++ b/cf_xarray/units.py
@@ -60,7 +60,7 @@ try:
 except ImportError:
     pass
 
-
+# ------
 # Reused with modification from MetPy under the terms of the BSD 3-Clause License.
 # Copyright (c) 2015,2017,2019 MetPy Developers.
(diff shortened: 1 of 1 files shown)


cf_xarray/units.py
```diff
@@ -83,13 +83,29 @@ def repl(m):
...
+units.define("year = 365.25 * day = yr")
+
+# Define commonly encountered units not defined by pint
+units.define("@alias degC = C = deg_C")
+units.define("@alias degK = deg_K")
+units.define("@alias day = d")
+units.define("@alias hour = h")  # Not the Planck constant...
 units.define(
-    "degrees_north = degree = degrees_N = degreesN = degree_north = degree_N = degreeN"
+    "degrees_north = 1 * degree = degrees_north = degrees_N = degreesN = degree_north = degree_N = degreeN"
```
dcherian, line 100: @aulemahal Why the `1 * `

aulemahal: I'm not sure about the `1 *` anymore... It might be unnecessary, after testing it seems to serve no purpose...

However, the third element is the string used in the abbreviated version of the unit, that you get with `f"{unit:~}"` (which we use here in `short_formatter`). It's to ensure the correct version is printed, instead of `"degrees_N"` or `"degrees_E"`.

aulemahal (approved):

keewis (commented):
looks good to me, although it feels a bit strange that the symbol of `degrees_north` should be `degrees_north`, I'd have assumed something like `°N` or `°` (same for `degrees_east`). I can't find anything on that in the CF conventions (not that I looked very hard), so maybe that's intentional? Why this is important is because something like
```python
u = ureg.Unit("degrees_north ** 2")
f"{u:~cf}"
```
would return `"degrees_north2"`, which is weird (unless I'm missing something?)

Actually, this is not really the fault of this PR: right now it returns `"degrees_N2"` which is not that much better.

Edit: now that I'm thinking about it, `"°N2"` or `"°2"` also look strange, so maybe we don't have much of a choice?

dcherian:
> I'd have assumed something like °N or ° (same for degrees_east)

I've realized that there are 2 distinct use-cases:
1. units for plotting where "fancy" stuff like °N would be great (basically emacs'  `prettify-symbols-mode` =) )
2. `.pint.dequantify().to_netcdf` where you want `degrees_north` to meet the standard.

dcherian:
The funny exponent syntax is from [UDUNITS](https://www.unidata.ucar.edu/software/udunits/udunits-2.2.28/udunits2lib.html#Syntax)

I see pint has  a  "pretty"  formatting mode. Is  there a way we  can pass  the "pretty" string?

```python
u = units.Unit("degrees_east")
f"{u:~P}"
```
```
degrees_east
```

PR #288, opened 2022-01-15
explicitly use the short formatter in the docs

keewis:
follow-up to #278. Since `pint` doesn't pass the modifier to the formatter at the moment, `":cf"` and `":~cf"` do the same thing right now. Despite that, it's better to add the `~` modifier when talking about the short format. 

Edit: see hgrecco/pint#1448


dcherian:
thanks!

PR #289, opened 2022-01-18
fix cf_xarray.utils._is_datetime_like

### LINKED ISSUE #285: Error when using add_canonical_attributes on 0 dimensional data / Bug in cf_xarray.utils._is_datetime_like
## Expected result
Using `da.cf.add_canonical_attributes()` on a 0 dimensional dataarray should return a new DataArray with the attributes added.

## What is happening
Due to the check on the data to check if this is a datetime, it raises an IndexError

## Reproducible code

```python
import xarray as xr
import cf_xarray
xr.DataArray(0, attrs={'standard_name':'sea_water_potential_temperature'}).cf.add_canonical_attributes()
```

Output:
```
IndexError                                Traceback (most recent call last)
/tmp/ipykernel_885592/3858241190.py in <module>
----> 1 xr.DataArray(0, attrs={'standard_name':'sea_water_potential_temperature'}).cf.add_canonical_attributes()

~/.local/share/virtualenvs/2021-11_obs_transition_zone-Sj4UnckL/lib/python3.8/site-packages/cf_xarray/accessor.py in add_canonical_attributes(self, override, skip, verbose, source)
   1802 
   1803                         # Don't add units to time variables (e.g., datetime64, ...)
-> 1804                         if key == "units" and _is_datetime_like(ds[var_name]):
   1805                             continue
   1806 

~/.local/share/virtualenvs/2021-11_obs_transition_zone-Sj4UnckL/lib/python3.8/site-packages/cf_xarray/utils.py in _is_datetime_like(da)
     19         import cftime
     20 
---> 21         if isinstance(da.data[0], cftime.datetime):
     22             return True
     23     except ImportError:

IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed
```

Indeed, `_is_datetime_like` checks the `da.data[0]` which does not work for 0 dimension data (nor for dimension >= 2):
```python
import cftime
from cf_xarray.utils import _is_datetime_like
_is_datetime_like(xr.DataArray([[cftime.datetime(2022, 1, 12)]]))
```
Returns False.

`cf_xarray.__version__` is 0.6.2

## Proposed solution

I can submit a PR with a fix and tests. I guess replacing `if isinstance(da.data[0], cftime.datetime):` in `_is_datetime_like` by
`if isinstance(da.data.reshape(-1)[0], cftime.datetime):` would be sufficient.

rcaneill:
Fixes #285 
I added 2 tests to verify this PR, they were not passed before and are now passed.
After a small benchmark, it seems that my proposed solution in #285 is not the fastest. I changed it to another similar solution.

### DIFF
--- a/cf_xarray/tests/test_accessor.py
+++ b/cf_xarray/tests/test_accessor.py
@@ -1416,12 +1416,12 @@ def test_add_canonical_attributes_0_dim():
 
 def test_datetime_like():
     """test for 0 or >= 2 time dimensions"""
-    import cftime
-
-    from cf_xarray.utils import _is_datetime_like
-
-    assert _is_datetime_like(xr.DataArray([[cftime.datetime(2022, 1, 12)]]))
-    assert _is_datetime_like(xr.DataArray(cftime.datetime(2022, 1, 12)))
+    da = xr.DataArray(
+        np.timedelta64(1, "D"),
+        attrs={"standard_name": "sea_water_age_since_surface_contact"},
+    )
+    new_attrs = da.cf.add_canonical_attributes().attrs
+    assert "units" not in new_attrs and "description" in new_attrs
 
 
 @pytest.mark.parametrize("override", [True, False])
--- a/cf_xarray/tests/test_accessor.py
+++ b/cf_xarray/tests/test_accessor.py
@@ -1419,8 +1419,8 @@ def test_datetime_like():
     import cftime
     from cf_xarray.utils import _is_datetime_like
 
-    assert _is_datetime_like(xr.DataArray([[cftime.datetime(2022, 1, 12)]])) == True
-    assert _is_datetime_like(xr.DataArray(cftime.datetime(2022, 1, 12))) == True
+    assert _is_datetime_like(xr.DataArray([[cftime.datetime(2022, 1, 12)]]))
+    assert _is_datetime_like(xr.DataArray(cftime.datetime(2022, 1, 12)))
 
 
 @pytest.mark.parametrize("override", [True, False])
--- a/cf_xarray/tests/test_accessor.py
+++ b/cf_xarray/tests/test_accessor.py
@@ -1417,6 +1417,7 @@ def test_add_canonical_attributes_0_dim():
 def test_datetime_like():
     """test for 0 or >= 2 time dimensions"""
     import cftime
+
     from cf_xarray.utils import _is_datetime_like
 
     assert _is_datetime_like(xr.DataArray([[cftime.datetime(2022, 1, 12)]]))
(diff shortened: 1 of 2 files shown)


rcaneill:
See [benchmark tests](https://nbviewer.ipython.org/gist/rcaneill/48d477a46d4353febd7a6e227a0ca2aa) for the executing times compared between different methods.

dcherian:
Thanks @rcaneill This function was copied from xarray,  would you mind sending the PR there (where it'll get a more thorough review)? Then we can copy over the fixed version.

I bet the bug gets triggered when trying to write a scalar cftime thing to file.

rcaneill:
Sure I can. I just tried to find the function in xarray but all my grep tests were unsuccessful... Do you know where precisely is was copied from?

dcherian:
Thanks for checking!

I don't know why we have our own version. Here's the xarray version, shall we just copy it over? Does that pass your tests?

https://github.com/pydata/xarray/blob/d3b6aa6d8b997df115a53c001d00222a0f92f63a/xarray/core/common.py#L1843-L1856

rcaneill:
I'll check it tomorrow!

May I ask why not importing the function from xarray instead of copying the code?

dcherian:
> May I ask why not importing the function from xarray instead of copying the code?

It's private, and so it could get renamed or moved and we'd have to keep up (low probability, I know...)

but the license allows us to copy so lets do that!

rcaneill:
xarray's function copied, passes all the tests.

See [benchmark](https://nbviewer.ipython.org/gist/rcaneill/7bf676fad09cd3a9f2f935fc2f73d5a9), there function is 100 times faster than naively testing the 1st element of the array for dask backend!
It is of same order of magnitude than the other solution for numpy backend.

cf_xarray/tests/test_accessor.py
```diff
@@ -1407,6 +1407,23 @@ def test_cf_standard_name_table_version():
...
+
+
+def test_datetime_like():
+    """test for 0 or >= 2 time dimensions"""
+    import cftime
+
+    from cf_xarray.utils import _is_datetime_like
+
+    assert _is_datetime_like(xr.DataArray([[cftime.datetime(2022, 1, 12)]]))
+    assert _is_datetime_like(xr.DataArray(cftime.datetime(2022, 1, 12)))
```
malmans2, line 1424: Maybe we could squeeze these tests, so we also avoid testing a private function? Something like:
```python
da = xr.DataArray(
    np.timedelta64(1, "D"),
    attrs={"standard_name": "sea_water_age_since_surface_contact"},
    )
new_attrs = da.cf.add_canonical_attributes().attrs
assert "units" not in new_attrs and "description" in new_attrs
```
What do you think?

rcaneill: Seems good for me

dcherian:
Thanks @rcaneill . Welcome to cf-xarray!

rcaneill:
I fixed a 2nd time the tests so they use cftime (as np timedelta was working previously). But it apears that I arrive 5 minutes too late... Should I open a new PR?

dcherian:
Ah thanks. Sure

PR #299, opened 2022-01-22
one more celsius alias

dcherian:


aulemahal (approved):
Seems good! The more aliases, the more fool-proof this gets!

PR #302, opened 2022-02-02
Fix unpickling by defining __setstate__

dcherian:
I think this is the problem
1.  __init__ isn't called during unpickling.
2. So self._obj is undefinited and goes through our _getattr machinery
   which requires self._obj as an argument, so we get a RecursionError

For some reason this is only (so far) triggered when stuffing xarray objects
in dask.delayed tasks and a distributed scheduler.

The test of pickling `da.cf` fails on main, so it would be good to fix
that in general.

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -933,6 +933,9 @@ class CFAccessor:
         self._obj = obj
         self._all_cell_measures = None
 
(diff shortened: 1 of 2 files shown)


dcherian:
I couldn't figure out how  to get `__new__` to work but `__setstate__` seems to work OK.

PR #315, opened 2022-03-25
new vertical coord decode and reordering

kthyng:
The vertical decoding has been added for `ocean_sigma_coordinate` and the free surface variable has been expanded dimensionally for all the calculations so that the results preserve the standard ordering of time, vertical, lat/Y, lon/X. Three of the tests do not work anymore, though. Expanding dimensionally apparently bumps the attributes off of s_rho, for one. For another, the values are very similar but do not match anymore — not sure why. Interested to see what @dcherian thinks!

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2190,7 +2190,7 @@ class CFDatasetAccessor(CFAccessor):
                 )
         return obj
 
-    def decode_vertical_coords(self, prefix="z", zname_in=None):
+    def decode_vertical_coords(self, *, outnames=None, prefix=None):
         """
         Decode parameterized vertical coordinates in place.
 
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2268,10 +2268,7 @@ class CFDatasetAccessor(CFAccessor):
                     terms["depth_c"] * terms["s"]
                     + (terms["depth"] - terms["depth_c"]) * terms["C"]
                 )
-                # # expand dims so ordering is preserved
-                # terms["eta"] = terms["eta"].expand_dims(
-                #     dim={terms["s"].name: terms["s"]}, axis=1
-                # )
+
                 # z(n,k,j,i) = S(k,j,i) + eta(n,j,i) * (1 + S(k,j,i) / depth(j,i))
                 ztemp = terms["eta"] * (1 + S / terms["depth"]) + S
 
@@ -2281,18 +2278,11 @@ class CFDatasetAccessor(CFAccessor):
                 S = (terms["depth_c"] * terms["s"] + terms["depth"] * terms["C"]) / (
                     terms["depth_c"] + terms["depth"]
                 )
-                # # expand dims so ordering is preserved
-                # terms["eta"] = terms["eta"].expand_dims(
-                #     dim={terms["s"].name: terms["s"]}, axis=1
-                # )
+
                 # z(n,k,j,i) = eta(n,j,i) + (eta(n,j,i) + depth(j,i)) * S(k,j,i)
                 ztemp = terms["eta"] + (terms["eta"] + terms["depth"]) * S
 
             elif stdname == "ocean_sigma_coordinate":
-                # # expand dims so ordering is preserved
-                # terms["eta"] = terms["eta"].expand_dims(
-                #     dim={terms["sigma"].name: terms["sigma"]}, axis=1
-                # )
                 # z(n,k,j,i) = eta(n,j,i) + sigma(k)*(depth(j,i)+eta(n,j,i))
                 ztemp = terms["eta"] + terms["sigma"] * (
                     terms["depth"] + terms["eta"]
@@ -2306,8 +2296,6 @@ class CFDatasetAccessor(CFAccessor):
             # add Axis attribute
             ztemp.attrs['axis'] = 'Z'
             ds.coords[zname] = ztemp
-            # reorder to time x depth x lat/Y x lon/X
-            ds.coords[zname].cf.transpose(*[dim for dim in ["T", "Z", "Y", "X"] if dim in ds.coords[zname].reset_coords(drop=True).cf.axes])
 
 
 @xr.register_dataarray_accessor("cf")
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2196,13 +2196,12 @@ class CFDatasetAccessor(CFAccessor):
 
         Parameters
         ----------
+        outnames : dict, optional
+            Keys of outnames are the input sigma/s coordinate variable name and
+            the values are the name to use for the associated vertical coordinate.
         prefix : str, optional
             Prefix for newly created z variables.
             E.g. ``s_rho`` becomes ``z_rho``
-        zname_in : str, optional
-            Name for new z variable, only makes sense to use if there is a single
-            variable being calculated. This is used in place of `prefix` if
-            provided.
 
         Returns
         -------
@@ -2233,11 +2232,14 @@ class CFDatasetAccessor(CFAccessor):
 
         allterms = self.formula_terms
         for dim in allterms:
-            suffix = dim.split("_")
-            if zname_in is None:
-                zname = f"{prefix}_" + "_".join(suffix[1:])
+            if prefix is None:
+                warnings.warn('`prefix` is being deprecated; use `outnames` instead.', DeprecationWarning)
+                # set outnames here
+                zname = outnames[dim]
+
             else:
-                zname = zname_in
+                suffix = dim.split("_")
+                zname = f"{prefix}_" + "_".join(suffix[1:])
 
             if "standard_name" not in ds[dim].attrs:
                 continue
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2233,15 +2233,12 @@ class CFDatasetAccessor(CFAccessor):
         allterms = self.formula_terms
         for dim in allterms:
             if prefix is None:
+                assert outnames is not None, 'if prefix is None, outnames must be provided'
                 # set outnames here
-                # need a default name option
-                if outnames is None:
-                    zname = f"z_{dim}"
-                else:
-                    try:
-                        zname = outnames[dim]
-                    except KeyError:
-                        print("Your `outnames` need to include a key of `dim`.")
+                try:
+                    zname = outnames[dim]
+                except KeyError:
+                    print("Your `outnames` need to include a key of `dim`.")
 
             else:
                 warnings.warn(
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2233,18 +2233,21 @@ class CFDatasetAccessor(CFAccessor):
         allterms = self.formula_terms
         for dim in allterms:
             if prefix is None:
-                warnings.warn(
-                    "`prefix` is being deprecated; use `outnames` instead.",
-                    DeprecationWarning,
-                )
                 # set outnames here
                 # need a default name option
                 if outnames is None:
                     zname = f"z_{dim}"
                 else:
-                    zname = outnames[dim]
+                    try:
+                        zname = outnames[dim]
+                    except KeyError:
+                        print('Your `outnames` need to include a key of `dim`.')
 
             else:
+                warnings.warn(
+                    "`prefix` is being deprecated; use `outnames` instead.",
+                    DeprecationWarning,
+                )
                 suffix = dim.split("_")
                 zname = f"{prefix}_" + "_".join(suffix[1:])
 
@@ -2295,8 +2298,6 @@ class CFDatasetAccessor(CFAccessor):
                     f"Coordinate function for {stdname!r} not implemented yet. Contributions welcome!"
                 )
 
-            # add Axis attribute
-            ztemp.attrs["axis"] = "Z"
             ds.coords[zname] = ztemp
 
 
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2265,12 +2265,12 @@ class CFDatasetAccessor(CFAccessor):
                     terms["depth_c"] * terms["s"]
                     + (terms["depth"] - terms["depth_c"]) * terms["C"]
                 )
-                # expand dims so ordering is preserved
-                terms["eta"] = terms["eta"].expand_dims(
-                    dim={terms["s"].name: terms["s"]}, axis=1
-                )
+                # # expand dims so ordering is preserved
+                # terms["eta"] = terms["eta"].expand_dims(
+                #     dim={terms["s"].name: terms["s"]}, axis=1
+                # )
                 # z(n,k,j,i) = S(k,j,i) + eta(n,j,i) * (1 + S(k,j,i) / depth(j,i))
-                ds.coords[zname] = terms["eta"] * (1 + S / terms["depth"]) + S
+                ztemp = terms["eta"] * (1 + S / terms["depth"]) + S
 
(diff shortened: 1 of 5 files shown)


kthyng:
Ah I see an immediate problem with my `zname_in` selection. It works for POM where there is one vertical coordinate but for FVCOM which also uses `ocean_sigma_coordinate`, there are two, and neither has `_` in them. So, `zname_in` should be a list I guess? or just a prefix on the full s coordinate name?

dcherian:
What are the variable names for FVCOM? I think the clearest solution might be

```
ds.cf.decode_vertical_coords(outnames={"s_rho": "z_rho", "s_w": "z_w"})
```

which means that we should deprecate `prefix` by first changing the default to None, then
```
if prefix is not None:
    warnings.warn(..., DeprecationWarning)
    # set outnames here
```

cf_xarray/accessor.py
```diff
@@ -2190,7 +2190,7 @@ def bounds_to_vertices(
                 )
         return obj
 
-    def decode_vertical_coords(self, prefix="z"):
+    def decode_vertical_coords(self, prefix="z", zname_in=None):
```
dcherian, line 2193:
```suggestion
    def decode_vertical_coords(self, *, outnames=None, prefix=None):
```

doc/whats-new.rst
```diff
@@ -2,6 +2,12 @@
 
 What's New
 ----------
+
+Development since release
+=========================
```
dcherian, line 7:
```suggestion
v 0.7.1 (unreleased)
====================
```

cf_xarray/accessor.py
```diff
@@ -2254,18 +2263,29 @@ def decode_vertical_coords(self, prefix="z"):
                     terms["depth_c"] * terms["s"]
                     + (terms["depth"] - terms["depth_c"]) * terms["C"]
                 )
+                # expand dims so ordering is preserved
+                terms["eta"] = terms["eta"].expand_dims(dim={terms["s"].name: terms["s"]}, axis=1)
```
dcherian, line 2267: I would prefer a more explicit transpose call at the end instead of expanding dimensions. Can you also transpose `romsds` in datasets for the tests.

kthyng: I tried to use the code I have for transposing but it didn't work because some dimensions were identified in the Dataset.dims but not in `cf-xarray` and I wasn't sure how to make it work generically. The code I was trying for that was (originally suggested by you, which I have in `xroms`), if the new depth variable is called `ztemp`:
```
ztemp.cf.transpose(*[dim for dim in ["T", "Z", "Y", "X"] if dim in ztemp.reset_coords(drop=True).cf.axes])
```

kthyng:
> What are the variable names for FVCOM? Can you also add an FVCOM test dataset please?

FVCOM won't work currently with xarray because the vertical coordinates have the same names as the dimensions and can't be read in without dropping the coordinate information, which is what has the vertical coordinate info. But I am trying to set up to be ready to do it when it works! 

> 
> I think the clearest solution might be a new `outnames` kwarg or maybe `decode_to`
> 
> ```
> ds.cf.decode_vertical_coords(outnames={"s_rho": "z_rho", "s_w": "z_w"})
> ```
> 
> which means that we should deprecate `prefix` by first changing the default to None, then
> 
> ```
> if prefix is not None:
>     warnings.warn(..., DeprecationWarning)
>     # set outnames here
> ```

I like this approach.

cf_xarray/accessor.py
```diff
@@ -2254,23 +2268,51 @@ def decode_vertical_coords(self, prefix="z"):
...
-                ds.coords[zname] = S + terms["eta"] * (1 + S / terms["depth"])
+                ztemp = terms["eta"] * (1 + S / terms["depth"]) + S
 
             elif stdname == "ocean_s_coordinate_g2":
                 # make sure all necessary terms are present in terms
                 # (depth_c * s(k) + depth(j,i) * C(k)) / (depth_c + depth(j,i))
                 S = (terms["depth_c"] * terms["s"] + terms["depth"] * terms["C"]) / (
                     terms["depth_c"] + terms["depth"]
                 )
+                # # expand dims so ordering is preserved
```
dcherian, line 2284: I don't see how this could work.

```
# S(k,j,i) + eta(n,j,i) * (1 + S(k,j,i) / depth(j,i))
```
 We want this to work even for subsets along a S-level (k=0) or time (n=0). In which case, inserting a dimension at axis=1 for eta would totally break (expand_dims with a scalar will use that scalar as size, or just error for negative value of s)

I experimented with explicit broadcasting but can't get that to work either.

```
# S(k,j,i) + eta(n,j,i) * (1 + S(k,j,i) / depth(j,i))
S = xr.DataArray(np.ones((4, 3, 2)), dims=("z", "y", "x"))
eta = xr.DataArray(np.ones((5, 3, 2)), dims=("t", "y", "x"))
depth = xr.DataArray(np.ones((3, 2)), dims=("y", "x"))
xr.broadcast(S, eta)
```

cf_xarray/accessor.py
```diff
@@ -2254,23 +2268,51 @@ def decode_vertical_coords(self, prefix="z"):
...
             else:
                 raise NotImplementedError(
                     f"Coordinate function for {stdname!r} not implemented yet. Contributions welcome!"
                 )
 
+            # add Axis attribute
+            ztemp.attrs["axis"] = "Z"
+            ds.coords[zname] = ztemp
+            # reorder to time x depth x lat/Y x lon/X
+            ds.coords[zname].cf.transpose(
```
dcherian, line 2308: If this is erroring it's probably because cf-xarray can't identify the T, Z, Y, X axes.

cf_xarray/accessor.py
kthyng: I think if we could count on dimensions being properly identified this would be straight-forward: the calculation could add singleton dimensions for any missing dimensions, do the calculation, the order would be proper, and then we could squeeze at the end to return. But, I don't think we can count on it working generally. So, I think we should remove all this attempt at reordering and just stick with the new calculation. I'll take it out and let me know if you think otherwise.

cf_xarray/accessor.py
kthyng: Yeah.

cf_xarray/accessor.py
dcherian: :+1: I can't see how we make it work in general without specific knowledge of the actual model variable names. Thanks for checking!

cf_xarray/tests/test_accessor.py
```diff
@@ -1036,7 +1037,9 @@ def test_param_vcoord_ocean_s_coord():
 
     romsds.s_rho.attrs["standard_name"] = "ocean_s_coordinate_g1"
     Zo_rho = romsds.hc * (romsds.s_rho - romsds.Cs_r) + romsds.Cs_r * romsds.h
-    expected = Zo_rho + romsds.zeta * (1 + Zo_rho / romsds.h)
+    # import pdb; pdb.set_trace()
```
dcherian, line 1040:
```suggestion
```

cf_xarray/tests/test_accessor.py
```diff
@@ -1036,7 +1037,9 @@ def test_param_vcoord_ocean_s_coord():
...
+    romsds["zeta"] = romsds.zeta.expand_dims(dim={"s_rho": romsds.s_rho}, axis=1)
+    expected = romsds.zeta * (1 + Zo_rho / romsds.h) + Zo_rho
```
dcherian, line 1042: revert this bit?

kthyng:
@dcherian Sorry we overlapped a bit just now. Could you take a look now? I think I have it correct but not positive.

cf_xarray/tests/test_accessor.py
```diff
@@ -1029,15 +1030,16 @@ def test_param_vcoord_ocean_s_coord():
         romsds.hc + romsds.h
     )
     expected = romsds.zeta + (romsds.zeta + romsds.h) * Zo_rho
-    romsds.cf.decode_vertical_coords()
+    romsds.cf.decode_vertical_coords(prefix="z")
```
dcherian, line 1033: Let's do `outnames` here.

cf_xarray/accessor.py
```diff
@@ -2223,12 +2227,22 @@ def decode_vertical_coords(self, prefix="z"):
...
         for dim in allterms:
-            suffix = dim.split("_")
-            zname = f"{prefix}_" + "_".join(suffix[1:])
+            if prefix is None:
+                warnings.warn(
+                    "`prefix` is being deprecated; use `outnames` instead.",
+                    DeprecationWarning,
+                )
+                # set outnames here
+                zname = outnames[dim]
```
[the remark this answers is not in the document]
kthyng: Ok I tried to do this with the default name behavior I also added. See what you think.

cf_xarray/accessor.py
```diff
@@ -2223,12 +2227,22 @@ def decode_vertical_coords(self, prefix="z"):
...
             "ocean_s_coordinate_g2": {"depth_c", "depth", "s", "C", "eta"},
+            "ocean_sigma_coordinate": {"sigma", "eta", "depth"},
         }
 
         allterms = self.formula_terms
         for dim in allterms:
-            suffix = dim.split("_")
-            zname = f"{prefix}_" + "_".join(suffix[1:])
+            if prefix is None:
+                warnings.warn(
```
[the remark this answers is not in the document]
kthyng: oops, fixed.

cf_xarray/datasets.py
```diff
@@ -16,6 +16,33 @@
...
+    {
+        "units": "sigma_level",
+        "long_name": "Sigma Stretched Vertical Coordinate at Nodes",
+        "positive": "down",
+        "standard_name": "ocean_sigma_coordinate",
+        "formula_terms": "sigma: sigma eta: zeta depth: depth",
+        "_CoordinateTransformType": "Vertical",
+        "_CoordinateAxisType": "GeoZ",
+        "_CoordinateZisPositive": "down",
+        "_CoordinateAxes": "sigma",
```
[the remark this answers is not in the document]
kthyng: I removed these — I think I just copied them in from an example dataset I was using.

cf_xarray/accessor.py
```diff
@@ -2254,23 +2268,33 @@ def decode_vertical_coords(self, prefix="z"):
...
                 # (depth_c * s(k) + depth(j,i) * C(k)) / (depth_c + depth(j,i))
                 S = (terms["depth_c"] * terms["s"] + terms["depth"] * terms["C"]) / (
                     terms["depth_c"] + terms["depth"]
                 )
+
                 # z(n,k,j,i) = eta(n,j,i) + (eta(n,j,i) + depth(j,i)) * S(k,j,i)
-                ds.coords[zname] = terms["eta"] + (terms["eta"] + terms["depth"]) * S
+                ztemp = terms["eta"] + (terms["eta"] + terms["depth"]) * S
+
+            elif stdname == "ocean_sigma_coordinate":
```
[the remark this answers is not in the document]
kthyng: I didn't quite follow this and didn't add anything. It looks like if eta and depth have certain names, then the resulting z coord should have a certain name, for `ocean_sigma_coordinate`, but I checked a dataset I'm using and the names didn't align so wasn't sure I was interpreting this correctly.

cf_xarray/accessor.py
```diff
@@ -2254,23 +2268,33 @@ def decode_vertical_coords(self, prefix="z"):
...
+                # z(n,k,j,i) = eta(n,j,i) + sigma(k)*(depth(j,i)+eta(n,j,i))
+                ztemp = terms["eta"] + terms["sigma"] * (terms["depth"] + terms["eta"])
 
             else:
                 raise NotImplementedError(
                     f"Coordinate function for {stdname!r} not implemented yet. Contributions welcome!"
                 )
 
+            # add Axis attribute
+            ztemp.attrs["axis"] = "Z"
```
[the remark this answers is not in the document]
kthyng: In the CIOFS ROMS output I am looking at (NOAA model), `s_rho` has `positive : up` in its attributes so it is identified as being a vertical coordinate. So that seems to make it more complicated for being able to label the z coordinate with `positive: up`. Or, maybe that label is found on the "s" coordinate in the formula, it can be removed from there and added to the resulting z coordinate?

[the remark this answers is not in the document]
dcherian: Ah that complicates things. Lets just skip adding attributes.

cf_xarray/accessor.py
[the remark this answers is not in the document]
dcherian: Ah ok. No worries then.

cf_xarray/accessor.py
```diff
@@ -2223,12 +2227,29 @@ def decode_vertical_coords(self, prefix="z"):
...
         allterms = self.formula_terms
         for dim in allterms:
-            suffix = dim.split("_")
-            zname = f"{prefix}_" + "_".join(suffix[1:])
+            if prefix is None:
+                # set outnames here
+                # need a default name option
+                if outnames is None:
+                    zname = f"z_{dim}"
+                else:
```
dcherian, line 2240:
```suggestion
```

Let's remove this and require `outnames` always when `prefix is None`

cf_xarray/tests/test_accessor.py
```diff
@@ -1056,6 +1058,17 @@ def test_param_vcoord_ocean_s_coord():
...
 
+def test_param_vcoord_ocean_sigma_coordinate():
+    expected = pomds.zeta + pomds.sigma * (pomds.depth + pomds.zeta)
+    pomds.cf.decode_vertical_coords(outnames={"sigma": "z"})
+    assert_allclose(pomds.z.reset_coords(drop=True), expected.reset_coords(drop=True))
+
+    copy = pomds.copy(deep=True)
+    del copy["zeta"]
+    with pytest.raises(KeyError):
+        copy.cf.decode_vertical_coords()
```
dcherian, line 1069:
```suggestion
        copy.cf.decode_vertical_coords()
        
    with pytest.raises(KeyError):
        copy.cf.decode_vertical_coords(outnames={})
```

dcherian (commented):
Looks good to me now. Just 2 v. minor comments. Thanks!

The test failures are unrelated.

cf_xarray/accessor.py
```diff
@@ -2223,12 +2227,28 @@ def decode_vertical_coords(self, prefix="z"):
...
-            zname = f"{prefix}_" + "_".join(suffix[1:])
+            if prefix is None:
+                assert (
+                    outnames is not None
+                ), "if prefix is None, outnames must be provided"
+                # set outnames here
+                try:
+                    zname = outnames[dim]
+                except KeyError:
+                    print("Your `outnames` need to include a key of `dim`.")
```
dcherian, line 2243:
```suggestion
                    raise KeyError("Your `outnames` need to include a key of `dim`.")
```

cf_xarray/tests/test_accessor.py
```diff
@@ -1056,6 +1058,20 @@ def test_param_vcoord_ocean_s_coord():
...
 
 
+def test_param_vcoord_ocean_sigma_coordinate():
+    expected = pomds.zeta + pomds.sigma * (pomds.depth + pomds.zeta)
+    pomds.cf.decode_vertical_coords(outnames={"sigma": "z"})
+    assert_allclose(pomds.z.reset_coords(drop=True), expected.reset_coords(drop=True))
+
+    copy = pomds.copy(deep=True)
+    del copy["zeta"]
+    with pytest.raises(KeyError):
```
dcherian, line 1068:
```suggestion
    with pytest.raises(AssertionError):
```

dcherian:
Thanks!

PR #329, opened 2022-04-19
correct docstring

### LINKED ISSUE #326: Error in docstring of vertices_to_bounds
There is an error in the docstring of `vertices_to_bounds` for the parameter `vertices`. The docstring says that it must be of shape (N, 2) or (N, M, 4), while in reality it must be (N+1,) (for the 1-D case, I guess (N+1, M+1) for the 2D case?).
Below is the line in the docstring (line 126).

https://github.com/xarray-contrib/cf-xarray/blob/88eff8342f92319a8a01cabddf029caf2bd8263b/cf_xarray/helpers.py#L113-L126

## Minimal example
```python
import numpy as np
import xarray as xr
import cf_xarray
print(cf_xarray.vertices_to_bounds(xr.DataArray(np.arange(5)), ('bounds', 'dim_0')))
```

prints
```
<xarray.DataArray (bounds: 2, dim_0: 4)>
array([[0, 1, 2, 3],
       [1, 2, 3, 4]])
Dimensions without coordinates: bounds, dim_0
```
which is what we expect.

## Proposed solution

I can correct the docstring in a PR. I however don't understand the 2D case, so if anyone can guide me on this I'll be happy.

rcaneill:
Fix #326

### DIFF
--- a/cf_xarray/helpers.py
+++ b/cf_xarray/helpers.py
@@ -123,7 +123,7 @@ def vertices_to_bounds(
     Parameters
     ----------
     vertices : DataArray
-        The bounds to convert. Must be of shape (N + 1) or (N + 1, M + 1).
+        The vertices to convert. Must be of shape (N + 1) or (N + 1, M + 1).
     out_dims : Sequence[str],
         The name of the dimension in the output. The first is the 'bounds'
         dimension and the following are the coordinate dimensions.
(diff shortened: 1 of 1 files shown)


aulemahal (approved):
Looks good to me, sorry for the initial mistake!

dcherian:
Thanks @rcaneill

PR #331, opened 2022-04-22
Check attrs before printing guess

### LINKED ISSUE #330: Should we only show guesses that matter?
Right now `guess_coord_axis(verbose=True)` prints all matches, even for variables that already have the appropriate attributes.
Should we only show the guesses that matter (i.e., the guess for variables with missing attributes)?

I think it gets even more confusing when multiple guesses are shown, such as this:
```
I think 'isobaricInhPa' is of type 'Z'. It matched re.compile('(z|nav_lev|gdep|lv_|[o]*lev|bottom_top|sigma|h(ei)?ght|altitude|depth|isobaric|pres|isotherm)[a-z_]*[0-9]*')
I think 'isobaricInhPa' is of type 'X'. It matched re.compile('x|i|nlon|ni')
``` 
The variable is correctly assigned to the Z axis, but I find the print a bit confusing.

malmans2:
Closes #330

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1732,7 +1732,7 @@ class CFAccessor:
         """
         obj = self._obj.copy(deep=True)
         for var in obj.coords.variables:
-            var_is_coord = var in sum(obj.cf.coordinates.values(), [])
+            var_is_coord = any(var in val for val in obj.cf.coordinates.values())
             if not var_is_coord and obj[var].ndim == 1 and _is_datetime_like(obj[var]):
                 if verbose:
                     print(
@@ -1742,12 +1742,12 @@ class CFAccessor:
                 continue  # prevent second detection
 
             for name, pattern in regex.items():
-                var_is_axis = var in sum(obj.cf.axes.values(), [])
-                var_is_coord = var in sum(obj.cf.coordinates.values(), [])
+                var_is_axis = any(var in val for val in obj.cf.axes.values())
+                var_is_coord = any(var in val for val in obj.cf.coordinates.values())
                 if (name in _AXIS_NAMES and var_is_axis) or (
                     name in _COORD_NAMES and var_is_coord
                 ):
-                    continue
+                    continue  # skip known axes/coordinates and prevent multiple guesses
                 # match variable names
                 if pattern.match(var.lower()):
                     if verbose:
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1743,10 +1743,10 @@ class CFAccessor:
                 continue  # prevent second detection
 
             for name, pattern in regex.items():
-                is_axis = var in sum(obj.cf.axes.values(), [])
-                is_coord = var in sum(obj.cf.coordinates.values(), [])
-                if (name in _AXIS_NAMES and is_axis) or (
-                    name in _COORD_NAMES and is_coord
+                var_is_axis = var in sum(obj.cf.axes.values(), [])
+                var_is_coord = var in sum(obj.cf.coordinates.values(), [])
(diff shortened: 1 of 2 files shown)


cf_xarray/accessor.py
```diff
@@ -1733,21 +1733,23 @@ def guess_coord_axis(self, verbose: bool = False) -> DataArray | Dataset:
...
                     print(
                         f"I think {var!r} is of type 'time'. It has a datetime-like type."
                     )
-                obj[var].attrs = dict(ChainMap(obj[var].attrs, ATTRS["time"]))
+                obj[var].attrs = new_attrs
                 continue  # prevent second detection
 
             for name, pattern in regex.items():
                 # match variable names
                 if pattern.match(var.lower()):
```
dcherian, line 1746: I think we can be a little more explicit. If it matches the pattern for `X` but the `axis` attribute exists, just skip.

malmans2: I changed it a little. The new checks avoid adding attributes to pre-defined axes and coordinates

dcherian: We could have a variable has the standard name "latitude", but not the "units" string. Currently this code would add the units. I think we should preserve that.

malmans2: isn't that `cf.add_canonical_attributes` though?

cf_xarray/accessor.py
```diff
@@ -1732,7 +1732,8 @@ def guess_coord_axis(self, verbose: bool = False) -> DataArray | Dataset:
         """
         obj = self._obj.copy(deep=True)
         for var in obj.coords.variables:
-            if obj[var].ndim == 1 and _is_datetime_like(obj[var]):
+            var_is_coord = var in sum(obj.cf.coordinates.values(), [])
```
dcherian, line 1735:
```suggestion
            var_is_coord = any(var in val for val in obj.cf.coordinates.values())
```

is this easier to read?

cf_xarray/accessor.py
```diff
@@ -1741,6 +1742,12 @@ def guess_coord_axis(self, verbose: bool = False) -> DataArray | Dataset:
                 continue  # prevent second detection
 
             for name, pattern in regex.items():
+                var_is_axis = var in sum(obj.cf.axes.values(), [])
+                var_is_coord = var in sum(obj.cf.coordinates.values(), [])
```
dcherian, line 1746:
```suggestion
```

Isn't this the same as above?

cf_xarray/accessor.py
dcherian: Good point!

I've been relying on `guess_coord_axis` to do it for me.


Let's add a notice to whats-new saying this "feature" has been removed.

cf_xarray/accessor.py
malmans2: yes, let's go for that one

cf_xarray/accessor.py
malmans2: yes, but we need to re-assess whether it's a coord in the for loop, otherwise multiple guesses are allowed.
I was undecided whether to copy and paste, create a function just for this, or maybe add a couple of private methods:
`obj.cf._var_is_coord(var)`

dcherian: Let's just add a comment. It's a single line that's pretty readable so no need for a function

cf_xarray/accessor.py
malmans2: done

cf_xarray/accessor.py
malmans2: done

dcherian:
Thanks @malmans2

PR #343, opened 2022-06-29
Explicit xarray imports

aulemahal:
In `accessor.py`, a few xarray classes are used without their module being explicitly imported. It seems that the latest commits to xarray have modified some import behaviour and it broke cf-xarray. Specifically, it seems that:
```
import xarray as xr
from xarray import DataArray, Dataset
from xarray.core.arithmetic import SupportsArithmetic
```
is not sufficient to import `xr.core.resample`. See https://github.com/pangeo-data/xESMF/runs/7119669567?check_suite_focus=true

This PR fixes the problem simply by explicitly importing those "wrapped classes".

I'm not sure if this should be considered a breaking change for xarray. If yes, then the fix might be best done there, but I haven't found exactly the source of the bug. It was a lot simpler to do what I did here.


dcherian:
Yes this looks better in general.

> I'm not sure if this should be considered a breaking change for xarray.

Me neither. @keewis do you have any thoughts here?

keewis:
Not sure. The aim of the PR that introduced this change (pydata/xarray#6702) was to improve the typing, so this was probably a unintended side-effect. We should probably open a issue to see what the others think?

PR #347, opened 2022-07-06
Correctly shaped bounds for add_bounds method

### LINKED ISSUE #346: wrongly shaped bounds from add_bounds method
Hey there,

was tinkering around with the `add_bounds` method and got this (from dump):
```bash
float yc_bounds(bounds, yc) ;
```

From the [CF-conventions](http://cfconventions.org/cf-conventions/cf-conventions.html#cell-boundaries), I found the following example for cell boundaries:
```bash
float A_bnds(eta, 2) ;
```
Here the shape is transposed.

Also `ncview` complains about the bounds with:
```
Error, compiled with max number of vertices for bounds var of 50!  But found a var with n=600
```

So I guess, the bounds should be transposed.

MuellerSeb:
Closes #346


cf_xarray/accessor.py
```diff
@@ -480,7 +480,7 @@ def _guess_bounds_dim(da, dim=None):
     first = (bounds.isel({dim: 0}) - diff.isel({dim: 0})).assign_coords(
         {dim: da[dim][0]}
     )
-    result = xr.concat([first, bounds], dim=dim)
+    result = xr.concat([first, bounds], dim=dim).transpose()
```
dcherian, line 483:
```suggestion
    result = xr.concat([first, bounds], dim=dim).transpose("bounds", ...)
```

This will make sure it's the first dimension

dcherian:
Thanks @MuellerSeb . Thanks for taking the time to contribute!

Can you modify the existing tests to assert that bounds is the first dimension?

To fix the tests, you'll have to transpose the "expected" values in the tests too.

MuellerSeb:
> I think you need to fix one more test with a transpose: ` def test_vertices_to_bounds():`

Was just wondering, why only one out of three asserts there fails.

dcherian:
@aulemahal can you take a look here please?

aulemahal:
Last commits should have fixed the last issues in the helpers as well as adapted the new 2D bounds with this change!

MuellerSeb:
LGTM. Thanks for your help!

PR #354, opened 2022-07-22
Add support for 'flag_masks'

### LINKED ISSUE #199: Support `flag_masks`
From http://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#flags

``` python
int basin(lat, lon);
       standard_name: region;
       flag_values: 1, 2, 3;
       flag_meanings:"atlantic_arctic_ocean indo_pacific_ocean global_ocean";
data:
   basin: 1, 1, 1, 1, 2, ..... ;
```

It'd be nice to provide `.cf.__eq__` and allow
``` python
ds.where(basin.cf == "atlantic_arctic_ocean")  # becomes ds.where(basin == 1)
```

or `ds.where(basin.cf.flag == "atlantic_arctic_ocean"))`?

We could do something similar for `.isin` too

I don't understand `flag_masks` but we could probably do something similar
``` python
  byte sensor_status_qc(time, depth, lat, lon) ;
    sensor_status_qc:long_name = "Sensor Status" ;
    sensor_status_qc:_FillValue = 0b ;
    sensor_status_qc:valid_range = 1b, 63b ;
    sensor_status_qc:flag_masks = 1b, 2b, 4b, 8b, 16b, 32b ;
    sensor_status_qc:flag_meanings = "low_battery processor_fault
                                      memory_fault disk_fault
                                      software_fault
                                      maintenance_required" ;
```

Descanonge:
Closes #199 

Now support CF flag variables encoding for mutually exclusive flags ('flag_values' attribute), independent flags ('flag_masks' attribute) or a mix of both (both attributes present).

- Change `create_flag_dict()`. The dictionnary values are tuples of both the 'flag_masks' and 'flag_values'. Either are set to None if not present in attributes.
- Add `extract_flags()` to retrieve boolean masks of any type of flag. Variable name is that of the 'flag_meaning' extracted. Can extract a single flag (returns a DataArray) or multiple ones (returns a Dataset).
- Change implementation of __eq__, __ne__, isin. All are thin wrappers around `extract_flags()`.
- Change __repr__ to list flags depending on their type (mutually exclusive or independent). Do not list the corresponding value anymore (it becomes complicated when mixing both types...).

It breaks rich comparisons. I do not understand in what situation they should be used. If it is necessary to keep them, they can be easily limited to work for mutually exclusive flags only.

Tests and documentation were not changed. If the implementation proposed is accepted, I can work on it as well.


dcherian:
Thanks @Descanonge I'll take a look soon.

dcherian:
> If it is necessary to keep them, they can be easily limited to work for mutually exclusive flags only.

Yes we should keep the rich comparisons. Those are the failing tests. `da.cf == "Altantic Ocean"` where `"Atlantic Ocean"` is in `flag_meanings`.

What do you think of adding a new `flags` property as discussed in #199,

Then `da.cf == "Atlantic Ocean"` and `da.cf.flags["Atlantic Ocean"]` would be identical. The latter syntax would generalize to flag masks too.

Descanonge:
I adapted the code for the interface proposed in the [issue](https://github.com/xarray-contrib/cf-xarray/issues/199#issuecomment-1198723719).
The `flags` property returns a Dataset containing all boolean masks. For instance this allows:
```
da.cf.flags.low_battery & ~da.cf.flags.maintenance_mode
```

I have kept rich comparisions so we can still do `da.cf == 'low_battery'` or `da.cf.isin(['low_battery'])`.

The dataset is lazily evaluated but not cached, so unless storing in a variable (`flags = da.cf.flags`) we add some overhead...

If the implementation is satisfactory I will fix the tests and docs.

cf_xarray/accessor.py
```diff
@@ -953,18 +953,36 @@ def __getattr__(self, attr):
         )
 
 
-def create_flag_dict(da):
+def create_flag_dict(da) -> Mapping[Hashable, Sequence]:
```
dcherian, line 956: Can we type the "Sequence" more explicitly please

cf_xarray/accessor.py
```diff
@@ -979,110 +997,6 @@ def __init__(self, obj):
...
-        Compare flag values against `other`.
-
-        `other` must be in the 'flag_meanings' attribute.
-        `other` is mapped to the corresponding value in the 'flag_values' attribute, and then
-        compared.
-        """
-        flag_dict = self._assert_valid_other_comparison(other)
-        return self._obj != flag_dict[other]
-
-    def __lt__(self, other):
```
dcherian, line 1012: Lets bring these other comparisons back. They coul be useful for quality levels for example

cf_xarray/datasets.py
```diff
@@ -412,7 +412,7 @@ def _create_inexact_bounds():
     )
 )
 
-
+# Same as flags_excl but easier to read
```
dcherian, line 415: I readded the test to make sure we have no regressions.

dcherian:
Sorry for the large delay here. This is a great contribution, and I'm excited to merge it.

 I have a few requests but think this is quite close. It would also be nice to add documentation but that could be in a future PR.

cf_xarray/accessor.py
```diff
@@ -1051,18 +1051,36 @@ def __getattr__(self, attr):
...
+    flag_values = da.attrs.get("flag_values", [None] * n_flag)
+    flag_masks = da.attrs.get("flag_masks", [None] * n_flag)
+
+    if not (n_flag == len(flag_values) == len(flag_masks)):
+        raise ValueError(
+            "Not as many flag meanings as values or masks. "
+            "Please check the flag_meanings, flag_values, flag_masks attributes "
+        )
+
+    return dict(zip(flag_meanings, zip(flag_masks, flag_values)))
```
dcherian, line 1083: It'd be nice to use a `namedtuple` with fields: `masks` and `meanings`

cf_xarray/accessor.py
```diff
@@ -1116,7 +1117,7 @@ def __lt__(self, other):
         compared.
         """
         flag_dict = self._assert_valid_other_comparison(other)
-        return self._obj < flag_dict[other]
+        return self._obj < flag_dict[other][1]
```
dcherian, line 1120: needs updating to namedtuple

cf_xarray/accessor.py
```diff
@@ -2744,22 +2745,128 @@ def __getitem__(self, key: Hashable | Iterable[Hashable]) -> DataArray:
...
+        Compare flag values against `other`.
+
+        `other` must be in the 'flag_meanings' attribute.
+        """
+        if other not in self.flags.data_vars:
+            raise ValueError(
+                f"Flag meaning {other!r} not in known meanings "
+                f"{list(self.flags.data_vars)!r}"
+            )
+        return self.flags[other].rename(self._obj.name)
```
dcherian, line 2759: I thought it was better to preserve the flag variable name since that's consistent with other comparisons

dcherian: Also  I think this is a big slowdown since we now loop over all flag_meanings for a single comparison. 

The old version was just a single pass through the data. Since we know what we're looking for, we should be just doing a single pass over the data.

cf_xarray/accessor.py
```diff
@@ -2744,22 +2745,128 @@ def __getitem__(self, key: Hashable | Iterable[Hashable]) -> DataArray:
...
+        return ~self.flags[other].rename(self._obj.name)
+
+    @property
+    def flags(self) -> Dataset:
+        """
+        Dataset containing boolean masks of available flags.
+        """
+        return self._extract_flags()
+
+    def _extract_flags(self, flags: Sequence[Hashable] | None = None) -> Dataset:
```
dcherian, line 2781: There is an alternative implementation here: https://nbviewer.org/gist/scottyhq/efd583d66999ce8f6e8bcefa81545b8d

```python
def extract_masks(da):
    ''' given a da.fmask return all mask arrays as an xarray dataset'''
    da = da.astype('uint8')
    dsM = da.to_dataset()
    
    explode = (da.shape[0], da.shape[1], 8)
    masks = np.unpackbits(da, axis=1).reshape(explode)
    
    aerosols = masks[:,:,0:2].sum(axis=-1)
    high_aerosol = (aerosols == 3)
    moderate_aerosol = (aerosols == 2)
    low_aerosol = (aerosols == 1)

    # Now to get a mask array it looks like this:
    masks = masks.astype('bool')
    water = masks[:,:,2]
    snow = masks[:,:,3]
    cloud_shadow = masks[:,:,4]
    cloud_adjacent = masks[:,:,5]
    cloud = masks[:,:,6]
    
    dsM['water'] = (('y','x'), water)
    dsM['snow'] = (('y','x'), snow)
    dsM['cloud_shadow'] = (('y','x'), cloud_shadow)
    dsM['cloud_adjacent'] = (('y','x'), cloud_adjacent)
    dsM['cloud'] = (('y','x'), cloud)
    dsM['high_aerosol'] = (('y','x'), high_aerosol)
    dsM['mod_aerosol'] = (('y','x'), moderate_aerosol)
    dsM['low_aerosol'] = (('y','x'), low_aerosol)
    dsM = dsM.drop_vars('Fmask')

    return dsM
```

for data

```
(LSB to MSB)

0: cirrus
1: cloud
2: cloud/shadow adjacent
3: cloud shadow
4: snow/isce
5: water
6-7: aerosol level
    - 11=high aerosol (3)
    - 10 moderate (2)
    - 01 low (1)
    - 00 climatrology aerosol (0)
```

Descanonge: I discarded np.unpackbits early when prototyping since it only works for 8 bits at most.
I am not sure if it is substantially faster than our current approach of applying a bitwise AND. The bitwise operation also directly allows for easy comparison when combining 'flag_values' and 'flag_masks', in contrast to here where additional computation is needed for aerosols (and where I think there is an error, low and moderate are not differentiated).
Completely unpacking bits is a nice approach, it's very clear to read and understand, but has some pitfalls I think.

Descanonge:
Thanks for the review, there are plenty of things to improve that's for sure. I am finishing writing my phd manuscript for the coming month so I am not super available but I'll try looking at it.

dcherian:
Ah no worries. Good luck with finishing your PhD manuscript!

cf_xarray/accessor.py
dcherian: Oh hmm.. that's a good point.

Can you link to some real dataset you use with this feature?

dcherian:
What do you think of these reprs?

### Plain text
image

image

image

image


### Rich

image

image

image

image

cf_xarray/accessor.py
```diff
@@ -2744,22 +2767,102 @@ def __getitem__(self, key: Hashable | Iterable[Hashable]) -> DataArray:
...
+        """
+        Return dataset of boolean mask(s) corresponding to `flags`.
+
+        Parameters
+        ----------
+        flags: Sequence[str]
+            Flags to extract. If empty (string or list), return all flags in
+            `flag_meanings`.
+        """
+        flag_dict = create_flag_dict(self._obj)
```
dcherian, line 2787: TODO cache this

PR #355, opened 2022-08-01
cf_xarray.datasets: hide auxiliary variables

mathause:
This hides all variables in `cf_xarray.datasets` that are not `xr.Dataset` such that `cf_xarray.datasets.<Tab>` is more helpful.


dcherian:
Great thanks!

PR #358, opened 2022-08-10
Performance improvement: Convert unnecessary (?) deep copies to shallow copies

lukelbd:
This converts several (unnecessary?) deep xarray copies to shallow copies throughout tests + `accessor.py`. I was surprised by the quite slow performance of this otherwise lightweight-seeming package and traced the issue to the copy calls.

As far as I can tell, this package only changes attribute dictionaries and adds new data/coordinate variables (i.e., does not touch the numpy arrays underlying each DataArray). If that's correct, `deep=True` should never be required, since both shallow and deep copies copy the various dictionaries associated with each xarray object (i.e. `data_vars`, `coords`, `attrs`, `encoding`). The deep version only differs in that it copies the (possibly enormous) underlying numpy arrays.

Here's a quick demo of "shallow" xarray copies (xarray version `2022.3.0`):

```python
# DataArray attributes within a DataArray copy
da1 = xr.DataArray([1, 2, 3], name='name', attrs={'foo': 'bar'})
da2 = da1.copy(deep=False)
da2.attrs.clear()
print(da1.attrs, da2.attrs)

# DataArray attributes within a Dataset copy
ds1 = da1.to_dataset()
ds2 = ds1.copy(deep=False)
ds2.name.attrs.clear()
print(ds1.name.attrs, ds2.name.attrs)
```

Result (attributes on the old version are unchanged):

```
{'foo': 'bar'} {}
{'foo': 'bar'} {}
```

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -1797,7 +1797,7 @@ class CFAccessor:
         DataArray or Dataset
             with appropriate attributes added
         """
-        obj = self._obj.copy(deep=True)
+        obj = self._obj.copy(deep=False)
         for var in obj.coords.variables:
             var_is_coord = any(var in val for val in obj.cf.coordinates.values())
             if not var_is_coord and obj[var].ndim == 1 and _is_datetime_like(obj[var]):
@@ -1937,7 +1937,7 @@ class CFAccessor:
         info, table, aliases = parse_cf_standard_name_table(source)
 
         # Loop over standard names
-        ds = self._maybe_to_dataset().copy()
+        ds = self._maybe_to_dataset().copy(deep=False)
         attrs_to_print: dict = {}
         for std_name, var_names in ds.cf.standard_names.items():
 
@@ -2209,7 +2209,7 @@ class CFDatasetAccessor(CFAccessor):
                 apply_mapper(_get_all, self._obj, key, error=False, default=[key])
             )
 
-        obj = self._maybe_to_dataset(self._obj.copy(deep=True))
+        obj = self._maybe_to_dataset(self._obj.copy(deep=False))
 
         bad_vars: set[str] = variables - set(obj.variables)
         if bad_vars:
@@ -2286,7 +2286,7 @@ class CFDatasetAccessor(CFAccessor):
         else:
             coords = keys
 
-        obj = self._maybe_to_dataset(self._obj.copy(deep=True))
+        obj = self._maybe_to_dataset(self._obj.copy(deep=False))
 
         for coord in coords:
             try:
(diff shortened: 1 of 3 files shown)


malmans2:
Thanks @lukelbd! This looks like a great improvement, let's see what @dcherian thinks about it.

In the meantime, could you please add a test to check that `cf-xarray` does not modify the original object?
(We might be testing it already, I didn't check)

cf_xarray/accessor.py
```diff
@@ -2209,7 +2209,7 @@ def add_bounds(self, keys: str | Iterable[str], *, dim=None):
                 apply_mapper(_get_all, self._obj, key, error=False, default=[key])
             )
 
-        obj = self._maybe_to_dataset(self._obj.copy(deep=True))
+        obj = self._maybe_to_dataset(self._obj.copy(deep=False))
```
dcherian, line 2212: Let's make sure the bounds aren't associated with the original dataset.

dcherian (commented):
Thanks @lukelbd  this is awesome!

I think I intended to verify the deep-copy of attrs, but never did and then forgot about it.

As @malmans2 suggested, can you make sure that the original object isn't modified for `add_canonical_attributes`, `add_bounds` `vertices_to_bounds` and `bounds_to_vertices` please (if we don't do already)

lukelbd:
No problem! Re: tests, I think this should do the trick. I wrote a helper function `_check_unchanged` that can compare initial, deep copies of testing Datasets/DataArrays against the version sent to the `cf` accessor methods, then I call that function at the end of `test_add_bounds`, `test_add_canonical_attributes`, and `test_bounds_to_vertices`. I use `id` comparison of attribute entries in case any of them are numpy arrays.

cf_xarray/accessor.py
lukelbd: I think this is now covered in `test_add_bounds`.

cf_xarray/tests/test_accessor.py
```diff
@@ -340,19 +340,18 @@ def test_getitem_ancillary_variables():
...
     del newair.lon.attrs["axis"]
     del newair.lat.attrs["axis"]
 
     renamed = popds.cf["TEMP"].cf.rename_like(newair)
+    assert original.TEMP.attrs["coordinates"] == "TLONG TLAT"
     for k in ["TLONG", "TLAT"]:
-        assert k not in renamed.coords
         assert k in original.coords
-    assert original.TEMP.attrs["coordinates"] == "TLONG TLAT"
+        assert k not in renamed.coords
```
dcherian, line 354: Nothing changed here. Correct?

lukelbd: Yep that's right. Didn't change it for any particularly big reason.

dcherian:
Thanks this is a great contribution!

Hope to see you here again soon. Feel free to open issues / PRs for things that would make your life easier.

PR #359, opened 2022-08-15
Accounting for NoneType coordinates in attrs/encoding

sol1105:
With https://github.com/pydata/xarray/pull/5514 it has become practice (at least for me) to occasionally disable the coordinates attribute to be written to file by setting `ds['var'].encoding['coordinates'] = None`. If used too generously, this may however lead to an AttributeError in the cf-xarray accessor:

>        if "coordinates" in obj.encoding:
>           search_in.update(obj.encoding["coordinates"].split(" "))
>E           AttributeError: 'NoneType' object has no attribute 'split'

I suggest this small update to keep that from happening.

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -254,12 +254,10 @@ def _get_axis_coord(obj: DataArray | Dataset, key: str) -> list[str]:
         )
 
     search_in = set()
-    coordinates = obj.encoding.get("coordinates", None)
+    attrs_or_encoding = ChainMap(obj.attrs, obj.encoding)
+    coordinates = attrs_or_encoding.get("coordinates", None)
     # Handles case where the coordinates attribute is None
     # This is used to tell xarray to not write a coordinates attribute
-    if coordinates:
-        search_in.update(coordinates.split(" "))
-    coordinates = obj.attrs.get("coordinates", None)
     if coordinates:
         search_in.update(coordinates.split(" "))
     if not search_in:
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -254,9 +254,9 @@ def _get_axis_coord(obj: DataArray | Dataset, key: str) -> list[str]:
         )
 
     search_in = set()
-    if "coordinates" in obj.encoding:
+    if "coordinates" in obj.encoding and obj.encoding["coordinates"]:
         search_in.update(obj.encoding["coordinates"].split(" "))
-    if "coordinates" in obj.attrs:
+    if "coordinates" in obj.attrs and obj.attrs["coordinates"]:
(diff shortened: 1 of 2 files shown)


cf_xarray/accessor.py
```diff
@@ -254,9 +254,9 @@ def _get_axis_coord(obj: DataArray | Dataset, key: str) -> list[str]:
         )
 
     search_in = set()
-    if "coordinates" in obj.encoding:
+    if "coordinates" in obj.encoding and obj.encoding["coordinates"]:
```
dcherian, line 257: Can we do 
```python
coordinates = obj.encoding.get("coordinates", None)
# Handles case where the coordinates attribute is None
# This is used to tell xarray to not write a coordinates attribute
if coordinates:
    ...
```

and similarly below.

sol1105: Thanks for your feedback. I tried to make the requested adjustments.

cf_xarray/accessor.py
```diff
@@ -254,10 +254,14 @@ def _get_axis_coord(obj: DataArray | Dataset, key: str) -> list[str]:
         )
 
     search_in = set()
-    if "coordinates" in obj.encoding:
-        search_in.update(obj.encoding["coordinates"].split(" "))
-    if "coordinates" in obj.attrs:
-        search_in.update(obj.attrs["coordinates"].split(" "))
+    coordinates = obj.encoding.get("coordinates", None)
```
dcherian, line 257:
```suggestion
	attrs_or_encoding = ChainMap(self._obj[name].attrs, self._obj[name].encoding)
    coordinates = attrs_or_encoding.get("coordinates", None)
```

cf_xarray/accessor.py
```diff
@@ -254,10 +254,14 @@ def _get_axis_coord(obj: DataArray | Dataset, key: str) -> list[str]:
...
+    # Handles case where the coordinates attribute is None
+    # This is used to tell xarray to not write a coordinates attribute
+    if coordinates:
+        search_in.update(coordinates.split(" "))
+    coordinates = obj.attrs.get("coordinates", None)
+    if coordinates:
+        search_in.update(coordinates.split(" "))
```
dcherian, line 264:
```suggestion
```

cf_xarray/accessor.py
sol1105: I was not sure if it was intentional that here a set of `attrs["coordinates"]` and `encoding["coordinates"]` is being constructed and not the ChainMap operator is being used (which respects then only the first mention of "coordinates" in attrs and encoding).

dcherian: Oh nice catch. I think we should treat coordinates in both attrs and encoding as a bug perhaps.

PR #361, opened 2022-08-18
check precision of vertex order

### LINKED ISSUE #360: order in `bounds_to_vertices`
Hey all,
thank for this, i greatly appreciate `cf_xarray`! I have one thing, since i work a lot with CORDEX curvilinear grids and 2D coordinates (and bounds). There is some cases in which computation of boundaries might have inprecise results, simply due to floating point limitations, e.g., very basic case i encountered:
```python
21.615 + 0.055 == 21.725 - 0.055  # e.g. right boundary of i-1 should be the same as left boundary of i
```
This would result in `False` which means that i can have neighbouring grid cells that do not share *excactly* the same boundary coordinate. In my case, i even transform the boundaries into a different CRS when i have 2D coordinates and when i order them *counterclockwise*, `bounds_to_vertices` with `order=None` will assum a *clockwise* order (the check for *counterclockwise* simply failes due to precision issues). For me this becomes important when using `xesmf` with conservative remapping, since [xesmf relies on "detecting" the order of 4D vertices](https://github.com/pangeo-data/xESMF/blob/d48285b5f65bd97820154f0253386ef69f65a36b/xesmf/frontend.py#L84) when using `bounds_to_vertices`.

I guess this could happen a lot if you have curvilinear CF datasets with boundary shapes of (N, M, 4). So i would suggest to change https://github.com/xarray-contrib/cf-xarray/blob/92ae91c17d8aae7654ec7469fead05bb4654c8f2/cf_xarray/helpers.py#L98

to something like this:

```python
order = "counterclockwise" if np.allclose(calc_bnds, values) else "clockwise" 
```
to allow for some tolerance... i have a more detailled gist of what i really do [here](https://gist.github.com/larsbuntemeyer/105d83c1eb39b1462150d3fabca0b66b). I am aware that this "order" is quite tricky to define, but I think it would be more a `cf_xarray` related issue than `xesmf`?

larsbuntemeyer:
closes #360


dcherian:
Can you add a test please

aulemahal (approved):
This change is good for me.

larsbuntemeyer:
Sorry for my laziness! I finally added a test where i check with a new dataset that has those precision issues in the bounds. Should i reduce the dimensions of the test data?

PR #370, opened 2022-10-21
2D bounds - simple version

### LINKED ISSUE #71: Update `add_bounds` to support 2D coordinate variables.
Should we change the name of this function?

Some relevant existing code is here: https://github.com/NCAR/pop-tools/blob/d771f8d7280696c890bb6a21255c5ad298c8b373/pop_tools/grid.py#L324-L363

cc @huard

aulemahal:
Fixes #71!

I went through `xgcm`'s code (using 0.6.1) and realized that the proposed solution for estimating 2D bounds was indeed quite simple! And that we already had the functions to implement it. 

So here I enable `ds.cf.add_bounds` to estimate 2D bounds by estimating the 1D bounds on the two axes successively, and then reordering the values to fit the conventions.

- The  `dim` argument can now be a list (of length 1 or 2)
- There is now a possibility that we add 2D bounds and 1D bounds on the same dataset. For now, if the `bounds` dim already exists and has a different length than the new one, the latter is renamed `bounds2`, which is a bit inelegant... Any ideas?
- I am always confused about the order to use... It doesn't really matter for most applications I guess (i.e. `xESMF`), but I'm not 100% sure I coded the right one.
- I will add more tests.

I first used @zxdawn's gist (https://gist.github.com/zxdawn/49c73e4d1ef9043580644e57eaa16d83) with xgcm 0.6.1. This implementation in cf-xarray does almost the exact same thing. I then compared with an old gist of mine (https://gist.github.com/aulemahal/58c0df9757635cb9b7f862a6d2a223ab) where I generated grid bounds for a RotatedPole grid. The idea there was to generate the 1D bounds on the regular rlat/rlon grid and then use cartopy/pyproj to project them to lat/lon.

I compared this PR's implementation with the projection one and the model-provided bounds for a regional model grid at a resolution of 0.11° and I got this :
image

The maps show the distance between the estimated bounds with the reference ones. Units are lat/lon degrees, so this means an error of maximum 2% of the grid resolution for this implementation. Seems reasonable to me, what do you think?

In the parent issue, there was a detailed proposition of moving the interesting grid and metric generation form xgcm to cf-xarray. I think it is still an interesting idea, but I felt the specific issue of 2D bounds was long overdue!

~CI issues seems to be unrelated to the PR, but rather to #365 and may be an update in xarray?~ Ok,there are indeed issues with this PR :sweat: .


aulemahal:
After discussion with a member of the team running the model from which I took my rotated pole grid example, I realized that the "reference" grid corners I was using were in fact generated in almost the same way as my "projection" method. So, disregard the left graphic above.

The right graphic still gives an idea of the error (in lat lon degrees) between the projection method (which we can treat as "exact") and the interpolation method implemented in this PR. The base grid above has a resolution of around 0.11°, thus the error goes up to 2% of that resolution.

This next figure is the same kind of comparison but with a 0.44° grid:
image
Here we see the error goes up to 0.05°, which is around 10%. This is not small.

But is it good enough for `cf-xarray` ?

My other idea for this issue was to:
- Implement a parsing for the `grid_mapping` CF data
- Generate the bounds of the 1D coordinates (`rlat` and  `rlon` for example)
- Project the bounds to lat/lon using `pyproj` or something similar. That's a new optional dependency though. It does come with `geopandas` so it should be common in our user's envs.
This should be way more precise. We could also have a switch between the two according to the presence of the needed dependencies and of a `grid_mapping` variable.

larsbuntemeyer:
@aulemahal as a user of `cf_xarray` and (CORDEX data), i would love to have it understand the `grid_mapping` attribute. I do the mapping quite regularly and i have also used `pyproj`, e.g., in this [prototype](https://github.com/euro-cordex/py-cordex/issues/16#issuecomment-1196078310). `pyproj` can actually understand the CF conventions regarding the `grid_mapping` attribute with the `CRS.from_cf` function. So that could probably also be useful to simply create not only bounds but also 2D coordinates in general if they are not part of the dataset. Would be nice if this could be done in `cf_xarray`...

dcherian:
> Here we see the error goes up to 0.05°, which is around 10%. This is not small.

Not small but the dataset should really contain these by default :)

@larsbuntemeyer Do you ever need something like this? If so,  does this approach look reasonable to you? 

> But is it good enough for cf-xarray ?

I really don't know. We should document this clearly though, both in the docstring and in the docs. Your example with the relative error graphic is nice. We should add that to the docs.

aulemahal:
@dcherian  Just pushed some doc in the `bounds` page and a few more lines in the docstrings.

I also checked if it works with `xESMF`. There's 2 lines to delete there, but otherwise, it seems to work as expected!

cf_xarray/accessor.py
```diff
@@ -2172,20 +2192,23 @@ def get_bounds_dim_name(self, key: str) -> str:
     def add_bounds(self, keys: str | Iterable[str], *, dim=None):
         """
         Returns a new object with bounds variables. The bounds values are guessed assuming
-        equal spacing on either side of a coordinate label.
+        equal spacing on either side of a coordinate label. The linear estimation is only a
+        coarse approximation, especially 2D bounds, like on curvilinear grids.
```
dcherian, line 2196:
```suggestion
        coarse approximation, especially 2D bounds on curvilinear grids. It is always better to use 
        bounds generated as part of the grid creation process.
```

cf_xarray/accessor.py
```diff
@@ -2226,9 +2249,10 @@ def add_bounds(self, keys: str | Iterable[str], *, dim=None):
             bname = f"{var}_bounds"
             if bname in obj.variables:
                 raise ValueError(f"Bounds variable name {bname!r} will conflict!")
-            obj.coords[bname] = _guess_bounds_dim(
-                obj[var].reset_coords(drop=True), dim=dim
-            )
+            out = _guess_bounds_dim(obj[var].reset_coords(drop=True), dim=dim)
+            if "bounds" in obj.dims and out.bounds.size != obj.bounds.size:
+                out = out.rename(bounds="bounds2")
```
dcherian, line 2254: Let's use  a `output_dim_name="bounds"` kwarg instead.

doc/bounds.md
```diff
@@ -10,3 +10,9 @@ See
 1. {py:func}`Dataset.cf.add_bounds`,
 1. {py:func}`cf_xarray.bounds_to_vertices`,
 1. {py:func}`cf_xarray.vertices_to_bounds`
+
+`cf_xarray` supports finding [coordinate bounds](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#cell-boundaries) as encoded in the CF `bounds` attribute. A useful feature for incomplete dataset is also the automatic bounds estimation possible through `cf.add_bounds`. This method will estimate the missing bounds by finding the middle points between elements of the given coordinate, but also by extrapolating to find the outer bounds of the grid. This linear estimation works well with rectilinear grids, but it is only a coarse approximation for curvilinear and simple irregular grids.
```
dcherian, line 14:
```suggestion
`cf_xarray` supports approximately estimating [coordinate bounds](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#cell-boundaries) as encoded in the CF `bounds` attribute. A useful feature for incomplete dataset is also the automatic bounds estimation possible through `cf.add_bounds`. This method will estimate the missing bounds by finding the middle points between elements of the given coordinate, but also by extrapolating to find the outer bounds of the grid. This linear estimation works well with rectilinear grids, but it is only a coarse approximation for curvilinear and simple irregular grids.
```

doc/bounds.md
```diff
@@ -10,3 +10,9 @@ See
...
+
+As an example, we took a "rotated pole" grid. It is defined on a rotated rectilinear grid which uses the `rlat` and  `rlon` 1D coordinates, over North America at a resolution of 0.44°. The datasets comes with 2D `lat` and `lon` coordinates. `cf_xarray` will estimate the bounds by linear interpolation (and extrapolation at the edges) of the existing `lon` and `lat`, which will yield good results on parts of the grid where the rotation is small, but the errors goes up in other places, as shown on the next figure which plots the distance (in lat-lon degrees) between the estimated bounds and the real ones.
```
dcherian, line 16:
```suggestion
As an example, we present a "rotated pole" grid. It is defined on a rotated rectilinear grid which uses the `rlat` and  `rlon` 1D coordinates, over North America at a resolution of 0.44°. The datasets comes with 2D `lat` and `lon` coordinates. `cf_xarray` will estimate the bounds by linear interpolation (extrapolation at the edges) of the existing `lon` and `lat`, which yields good results on parts of the grid where the rotation is small. However the errors is larger in other places, as seen when visualizing the distance in degrees between the estimated bounds and the true bounds.
```

doc/bounds.md
aulemahal: I meant that "finding" to be about `cf.get_bounds`. How's "parsing" ?

dcherian:
LGTM. Thanks a lot @aulemahal ! 

cc @jbusecke

jbusecke:
Oh this is dope! If I understand correctly what is implemented here (just took a very quick look, will test next week hopefully) this might help a lot with getting rid of the dreaded `autogenerate` logic in xgcm and pave the way to work much easier with gridded datasets, which often only provide the cell center coordinates. In that case IMO this 'is enough' because its better than not working with the dataset!

I am still very curious about the suggestions in @larsbuntemeyer s answer (e.g. an even more precise way to 'generate curvilinear grids' but I am not sure that is within the scope of cf-xarray? Maybe it is. Happy to talk more about this in the future. 

But mostly a huge thanks to @aulemahal for working on this!

PR #376, opened 2022-11-29
Fix add_bounds for heavily curved grids

### LINKED ISSUE #374: Guessing bounds on unindexed dimensions unsupported
Code says:
https://github.com/xarray-contrib/cf-xarray/blob/3795af6cb6f5d37ffac7e2a67f5b4ead1255b202/cf_xarray/accessor.py#L497-L500

That makes sense if you are trying to get the bounds of a dimension which has no index. In that case, xarray will create a range-like index out of thin air and you'll get bounds that do not correspond to anything present in the dataset.

However, a coordinate could be given along unindexed dimensions. The unindexed dimension itself is never used in itself,  it's only that the `diff` is taken along it.

Of course, that doesn't come up often with 1D coordinates, but it is quite frequent with 2D ones! I encountered this issue while trying to enable the new 2D bounds in `xESMF` . Sadly, I didn't detect it in my tests, as I was using rotated pole data, which almost always has `rlat` and `rlon`. This is not the case of other curvilinear grid types.

If I am right, then this error can't be detected in that function. We have to detect it before "extracting" the coordinate from the dataset, so within `cf.add_bounds`.

aulemahal:
Closes #374 
but also another issue that I bumped into while trying to update xESMF with the new 2D bounds estimation.

## Heavily curved grids
The previous method, implemented in #370, was creating bounds for each cell individually. This meant that on "heavily curved" grids the shared corners of two adjacent cells could have quite different values. Enough so that the  orientation check in `bounds_to_vertices` would fail. This, it would return vertices as if the bounds were stored in the "clockwise" order, instead of the correct "counterclockwise". I'm guessing that this wouldn't have been the only problem, but it's the one that forbid using it in `xESMF` where `bounds_to_vertices` is needed and the order can't be guessed easily.

Let me show an example.
```python3
import xarray as xr
import numpy as np
from shapely.geometry import Polygon
import matplotlib.pyplot as plt
import geopandas as gpd

ds = xr.tutorial.open_dataset("ROMS_example.nc", chunks={"ocean_time": 1})
ds.lat_rho.attrs['units'] = 'degrees_north'  # To detect the coordinates as the lat and lon to use
ds.lon_rho.attrs['units'] = 'degrees_west'
# To avoid #374, it is fixed by this PR too.
ds = ds.assign_coords(eta_rho=ds['eta_rho'], xi_rho=ds['xi_rho'])
ds = ds.cf.add_bounds(['lat_rho', 'lon_rho'])

lonb = ds.lon_rho_bounds.values
latb = ds.lat_rho_bounds.values

# Construct rectangle polygons from the bounds
ps = []
for i in range(191):
    for j in range(317):
        ps.append(
            Polygon([(lonb[i, j, k], latb[i, j, k]) for k in range(4)])
        )
gdf = gpd.GeoDataFrame(geometry=ps)
gdf.plot(edgecolor='k', alpha=0.5)
```
The figure below is a zoom on the result using cf_xarray at main:
image

The issue is: the 4 corners within the red circle are expected to be _the same point_. 
The solution : average together the 4 results we get for each output point.

To manage the edges, that required some acrobatics, but it seems to work well. The same code, run with this branch gives:
image
Much better.

## Adding bounds without indexes
I realized that this was not a real issue. The bounds guessing performs a _diff_ along the given dimension, so it doesn't really matter that this dimension has no index.  However, it does matter in a technical way, as it is used by xarray to align the diff output with the original da. This means, we can simply create a dummy index if one is missing and remove it before returning. This is what I implemented.

The only case where we shouldn't do that, is when someones requests bounds on a dimension that has no index. But this case is already covered by an earlier check in `add_bounds` since such a "dimension" doesn't corresponds to a variable.


cf_xarray/accessor.py
```diff
@@ -481,23 +482,80 @@ def _guess_bounds_dim(da, dim=None, out_dim="bounds"):
...
-            return xr.concat(
-                [
-                    daXY.isel(Xbnds=0, Ybnds=0),
-                    daXY.isel(Xbnds=0, Ybnds=1),
-                    daXY.isel(Xbnds=1, Ybnds=1),
-                    daXY.isel(Xbnds=1, Ybnds=0),
-                ],
-                out_dim,
-            ).transpose(..., "bounds")
+            # At this point, we might have different corners for adjacent cells, we average them together to have a nice grid
```
dcherian, line 485: lets stick  in its own helper function (`_guess_bounds_2d`)

cf_xarray/accessor.py
```diff
@@ -481,23 +482,80 @@ def _guess_bounds_dim(da, dim=None, out_dim="bounds"):
...
+                        ),
+                        axis=0,
+                    ),
+                ),
+                axis=-1,
+            )
+            return xr.DataArray(daXYm, dims=(*dim, "bounds"), coords=da.coords)
         else:
             dim = dim[0]
     if dim not in da.dims:
```
dcherian, line 551: I think we shoulld add the 1D code to its own `_guess_bounds_1d` helper function.

dcherian:
Thanks this seems like a good improvement, and really illustrates that datasets should be distributed with appropriate grid info!

Can you add your very nice figure to the docs too?

cf_xarray/accessor.py
aulemahal: Done! I also sent the functions to `helpers.py`.

dcherian:
OK LGTM!

Can you add a test? I think just checking a couple of rectangles in the figure would work.

aulemahal:
I added a test!
Instead of creating a new `ROMS` dataset from the one in xarray's data, I used `rotds`, which we already had. Thus I am indeed testing that  `cf_xarray` _does not_ generate bounds with the imprecision issue raised in #360. 

So, we are allowing these issues in `bounds_to_vertices` but not in the bounds we generate with `cf.add_bounds`.

dcherian:
> I added a test!

To clarify, there is an existing test? Or did you add a new one (it isn't in the diff)

aulemahal:
oops, I forgot to push! Here it is.

cf_xarray/tests/test_accessor.py
```diff
@@ -828,8 +841,9 @@ def test_add_bounds_nd_variable():
     actual = ds.cf.add_bounds("z", dim="x").z_bounds.reset_coords(drop=True)
     xr.testing.assert_identical(expected.transpose(..., "bounds"), actual)
 
-    with pytest.raises(NotImplementedError):
-        ds.drop_vars("x").cf.add_bounds("z", dim="x")
+    # Requesting bounds on a non-variable dimension
+    with pytest.raises(ValueError):
+        ds.drop_vars("x").cf.add_bounds("x")
```
dcherian, line 846: Last request: Can we raise a nice error message for this case?

dcherian (approved):

PR #379, opened 2022-12-05
allow custom criteria for name of DataArray

### LINKED ISSUE #378: "name" custom criteria not working for DataArray
```python
import cf_xarray
import xarray as xr

da = xr.DataArray([1, 2], coords={"lat": [3, 4]})
lat_criteria = {"latitude": {"name": "lat"}}
cf_xarray.set_options(custom_criteria=lat_criteria)
da.cf["latitude"]
```

Raises

```python-traceback
File ~/code/cf-xarray/cf_xarray/accessor.py:223, in _get_custom_criteria(obj, key, criteria)
    219                 results.update((var,))
    220             # also check name specifically since not in attributes
--> 223             elif criterion == "name" and re.match(patterns, var):
    224                 results.update((var,))
    225 return list(results)

File ~/.conda/envs/cf_xarray_dev/lib/python3.11/re/__init__.py:166, in match(pattern, string, flags)
    163 def match(pattern, string, flags=0):
    164     """Try to apply the pattern at the start of the string, returning
    165     a Match object, or None if no match was found."""
--> 166     return _compile(pattern, flags).match(string)

TypeError: expected string or bytes-like object, got 'ReprObject'
```

because the DataArray is converted to a temporary Dataset which uses a `ReprObject` as its name. An easy fix is to call `str` on `var` here:

https://github.com/xarray-contrib/cf-xarray/blob/3795af6cb6f5d37ffac7e2a67f5b4ead1255b202/cf_xarray/accessor.py#L217

mathause:
- [x] Closes #378

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -214,7 +214,7 @@ def _get_custom_criteria(
                 if re.match(patterns, obj[var].attrs.get(criterion, "")):
                     results.update((var,))
                 # also check name specifically since not in attributes
-                elif criterion == "name" and re.match(patterns, str(var)):
+                elif criterion == "name" and isinstance(var, str) and re.match(patterns, var):
                     results.update((var,))
     return list(results)
 
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -214,7 +214,7 @@ def _get_custom_criteria(
                 if re.match(patterns, obj[var].attrs.get(criterion, "")):
                     results.update((var,))
                 # also check name specifically since not in attributes
-                elif criterion == "name" and re.match(patterns, var):
+                elif criterion == "name" and re.match(patterns, str(var)):
                     results.update((var,))
     return list(results)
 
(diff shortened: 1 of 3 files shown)


cf_xarray/accessor.py
```diff
@@ -214,7 +214,7 @@ def _get_custom_criteria(
                 if re.match(patterns, obj[var].attrs.get(criterion, "")):
                     results.update((var,))
                 # also check name specifically since not in attributes
-                elif criterion == "name" and re.match(patterns, var):
+                elif criterion == "name" and re.match(patterns, str(var)):
```
dcherian, line 217: isn't this wrong when `var` is a Hashable like a tuple?

We could add a `and not isinstance(var, ReprObject)` instead. I think it's ok to depend on that private class

mathause: maybe `and isinstance(var, str)` for now.

dcherian: Sure that works for me

dcherian, line 217:
```suggestion
                elif criterion == "name" and isinstance(var, str) and re.match(patterns, var):
```

mathause:
Thanks!

PR #386, opened 2022-12-19
Fix for shapely 2.0

aulemahal:
This is a small fix to support  `shapely` 2.0. Changes were made between 1.8 and 2 that removed the implicit array interface the previous  `points_to_cf` was using.  We now need a check for  "MultiPoints", but the change still is minimal.
This works with the previous shapely too.


cf_xarray/geometry.py
```diff
@@ -195,7 +195,10 @@ def points_to_cf(pts: Union[xr.DataArray, Sequence]):
 
     x, y, node_count, crdX, crdY = [], [], [], [], []
     for pt in pts:
-        xy = np.atleast_2d(np.array(pt))
+        if hasattr(pt, "geoms"):  # MultiPoint
```
dcherian, line 198: Can we  do a `isinstance` check instead and a comment about shapely  2.0?

PR #394, opened 2023-01-24
Better typing

dcherian:


kthyng:
@dcherian I'm always trying to keep up to date with CI stuff — why change from precommit to a ci.yaml for mypy?

dcherian:
It's nicer for new contributors, who may not know how to handle mypy errors.

In Xarray, we did it to typecheck over multiple python versions, which is useful because the typing stuff keeps changing.

PR #408, opened 2023-01-30
Using regex package for match

kthyng:
The built-in re package does not allow for global flags like "(?i)" to be anywhere but the start of a pattern string now. The package `regex` still allows this, so it is optionally used for the `match` function if available.
* test added
* updating whats new


kthyng:
@dcherian How do I set up a test that will work for both with the optional dependences and without?

Also I wasn't sure which environment file to add `regex` to.

cf_xarray/accessor.py
```diff
@@ -3,6 +3,11 @@
 import functools
 import inspect
 import itertools
+
+try:
+    from regex import match
+except ImportError:
+    from re import match
 import re
```
dcherian, line 11:
```suggestion
    from re import match
```

cf_xarray/tests/test_accessor.py
```diff
@@ -1584,6 +1584,14 @@ def test_custom_criteria() -> None:
         ds["temperature"] = ("dim", np.arange(10))
         assert_identical(ds.cf["temp"], ds["temperature"])
 
+    # test that having a global regex expression flag later in the expression will work if
+    # regex is found
+    vocab = {"temp": {"name": "tem|(?i)temp"}}
+    ds = xr.Dataset()
+    ds["Tempblah"] = [0, 1, 2]
+    with cf_xarray.set_options(custom_criteria=vocab):
+        assert_identical(ds.cf["temp"], ds["Tempblah"])
```
dcherian, line 1593: Add  one of these for regex:
https://github.com/xarray-contrib/cf-xarray/blob/0f92bc282176df0427717663a05e8869d7822eff/cf_xarray/tests/__init__.py#L69

Then

```suggestion
	if has_regex:
       	 assert_identical(ds.cf["temp"], ds["Tempblah"])
       else:
       	with pytest.raises(...):
       		ds.cf["temp"]
```

dcherian, line 1593: Add  one of these for regex:
https://github.com/xarray-contrib/cf-xarray/blob/0f92bc282176df0427717663a05e8869d7822eff/cf_xarray/tests/__init__.py#L69

Then put this in a new test as
```python
@requires_regex
def test_regex_...():
    pass
```

dcherian:
Can you add some docs under custom criteria too please? Also we should regex as an optional dependency in `pyroject.toml`

cf_xarray/accessor.py
kthyng: `re` is also used on Line 2545. Would you prefer to treat `re.sub` to the same change as `re.match`? I didn't change it since it wasn't necessary.

dcherian: Oh didn't realize that. Let's keep as-is for now then.

dcherian:
You'll need to add `regex` to `ci/environment.yml` too.

cf_xarray/tests/test_accessor.py
kthyng: Have added this.

kthyng:
@dcherian I looked at the custom criteria docs and they are surprisingly good! Any specific topics you want me to add?

dcherian:
Just that installing `regex` lets you do more complicated things

kthyng:
I can demo a more complicated custom criteria — I have been using a similar set up for awhile though, it's just that `re` deprecated part of what made it possible, so it's not a new thing.

Any idea for how to address the mypy error? My idea let to the same error as before.

doc/whats-new.rst
```diff
@@ -3,6 +3,11 @@
 What's New
 ----------
 
+v0.7.8 (unreleased)
+===================
+
+- Change to ``accessor.py`` to continue supporting global flags in regular expressions that are not at start of pattern. (:pr:`408`). By `Kristen Thyng`_
```
dcherian, line 9:
```suggestion
- Optionally use the `regex` package to continue supporting global flags in regular expressions that are not at start of pattern. (:pr:`408`). By `Kristen Thyng`_
```

doc/custom-criteria.md
```diff
@@ -102,3 +102,17 @@ cfxr.set_options(custom_criteria=salt_criteria)
 
 ds.cf[["salinity"]]
 ```
+
+Here is an example of a more complicated custom criteria, which requires the package `regex` to be installed since a behavior (allowing global flags like "(?i)" for matching case insensitive) was recently deprecated in the `re` package. The custom criteria, called "vocab", matches – case insensitive – to the variable alias "sea_ice_u" a variable whose name includes "sea" and "ice" and "u" but not "qc" or "status", or "sea" and "ice" and "x" and "vel" but not "qc" or "status".
+
+```{code-cell}
```
dcherian, line 108: ideally this will fail and you'll need to add `regex` to `ci/docs.yml`

Let's also link to the `regex` documentation somewhere

kthyng: I think it ran because I *think* it is still a deprecation warning with Python 3.8 but becomes an error with Python 3.9 and the docs are on 3.8. But I did check the RTD build log and I don't see the deprecation warning in them. Still I'll go add `regex` to the docs env and link to the regex package docs.

kthyng: Ok I just added a commit with these items.

doc/custom-criteria.md
```diff
@@ -102,3 +102,18 @@ cfxr.set_options(custom_criteria=salt_criteria)
 
 ds.cf[["salinity"]]
 ```
+
+Here is an example of a more complicated custom criteria, which requires the package [`regex`](https://github.com/mrabarnett/mrab-regex) to be installed since a behavior (allowing global flags like "(?i)" for matching case insensitive) was recently deprecated in the `re` package. The custom criteria, called "vocab", matches – case insensitive – to the variable alias "sea_ice_u" a variable whose name includes "sea" and "ice" and "u" but not "qc" or "status", or "sea" and "ice" and "x" and "vel" but not "qc" or "status".
```
dcherian, line 106:
```suggestion
## More complex matches with `regex`

Here is an example of a more complicated custom criteria, which requires the package [`regex`](https://github.com/mrabarnett/mrab-regex) to be installed since a behavior (allowing global flags like "(?i)" for matching case insensitive) was recently deprecated in the `re` package. The custom criteria, called "vocab", matches – case insensitive – to the variable alias "sea_ice_u" a variable whose name includes "sea" and "ice" and "u" but not "qc" or "status", or "sea" and "ice" and "x" and "vel" but not "qc" or "status".
```

dcherian (approved):

dcherian:
LGTM. Thanks!

doc/custom-criteria.md
```diff
@@ -102,3 +102,22 @@ cfxr.set_options(custom_criteria=salt_criteria)
 
 ds.cf[["salinity"]]
 ```
+
+## More complex matches with `regex`
+
```
dcherian, line 107:
```suggestion
```

kthyng:
@dcherian should I do a next step on this or will you?

Also would it be possible to have a subsequent new release associated with this so I can use it in packages that use `cf-xarray`?

Issue #23, opened 2020-06-12
should 'X', 'Y' be synonyms for 'longitude', 'latitude'?


dcherian:
I am very unsure about this. Originally, they were synonyms. In 21c03876e1fba1aa94197dbf8c9ff1a8ffc955af I undid that.

Here is the current state of affairs
https://github.com/xarray-contrib/cf-xarray/blob/21c03876e1fba1aa94197dbf8c9ff1a8ffc955af/cf_xarray/accessor.py#L24-L89

Note that `'Z', 'T'` are synonymous with `'vertical', 'time'`.

This would be a good place for someone to chime in.

jthielen:
Based on what we've found in MetPy and in my own frequent use-case of gridded atmospheric data in projected spaces, I would argue that 'X', 'Y' should not be synonyms for 'longitude', 'latitude' in general. It is not uncommon that an 'X'/'Y' axis is present as a coordinate variable (CF terminology)/dimension coordinate (xarray terminology), but they are not 'longitude'/'latitude' (which is instead a 2D auxiliary coordinate variable/non-dimension coordinate). Both these types of coordinates are separately useful, and so, should ideally be available independently (although, in the special case of the plate carrée/`latitude_longitude` projection, they would be synonyms). The [Independent Latitude, Longitude, Vertical, and Time Axes](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#_independent_latitude_longitude_vertical_and_time_axes) and [Two-Dimensional Latitude, Longitude, Coordinate Variables](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#_two_dimensional_latitude_longitude_coordinate_variables) examples in the CF Conventions were most helpful for me in interpreting these. Elsewhere in the conventions, it at times seems implied that axes of 'X', 'Y' *are* synonyms for 'longitude', 'latitude', but that is clearly not the case based on these two examples. I wish the conventions were clearer in regards to these non-longitude/latitude horizontal coordinate variables/dimension coordinates.

For some background, these were quasi-synonymous in MetPy (i.e. 'X' could access 'longitude' even if >1D, but would default to a suitable dimension coordinate if present), but that was changed with https://github.com/Unidata/MetPy/issues/1090 / https://github.com/Unidata/MetPy/pull/1236. I've been quite happy with how MetPy's been handling it since the change.

dcherian:
[...]

PR #426, opened 2023-02-27
identify `X` and `Y` in rotated pole grids

### LINKED ISSUE #425: identify `rlon`/`rlat` as `X`/`Y`
It would be useful if we could have `rlon`, `rlat` also be identified as the `X`, `Y` axis from the variable name, e.g., adding them to the regex here?

https://github.com/xarray-contrib/cf-xarray/blob/ce9cd5a303e7344d55f9f85d49892d498a9cfbf2/cf_xarray/criteria.py#L139

and maybe also add `grid_longitude` and `grid_latitude` to the `X`, `Y` criteria here?

https://github.com/xarray-contrib/cf-xarray/blob/ce9cd5a303e7344d55f9f85d49892d498a9cfbf2/cf_xarray/criteria.py#L97

larsbuntemeyer:
closes #425 

This PR updates criteria and regex for identifying `X` and `Y` axis in rotated pole grids.

### DIFF
--- a/cf_xarray/criteria.py
+++ b/cf_xarray/criteria.py
@@ -94,14 +94,22 @@ coordinate_criteria: MutableMapping[str, MutableMapping[str, Tuple]] = {
         "positive": ("up", "down"),
     },
     "X": {
-        "standard_name": ("projection_x_coordinate",),
+        "standard_name": (
+            "projection_x_coordinate",
+            "grid_longitude",
+            "projection_x_angular_coordinate",
+        ),
         "_CoordinateAxisType": ("GeoX",),
         "axis": ("X",),
         "cartesian_axis": ("X",),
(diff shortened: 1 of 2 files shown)


dcherian:
Are any of these always 1D? Maybe the projection coordinates are? 

Basically we think that `axis` varibales should be 1D coordinate variables. This provides an easy way to refer to dimensions, as opposed to nD coordinate variables, and that is useful for array analysis.

So we don't usually set `axis="X"` on `longitude` variables because they are frequently 2D.

See these discussions:
https://github.com/xarray-contrib/cf-xarray/issues/23
https://github.com/xarray-contrib/cf-xarray/issues/87

dcherian:
Looks great!

larsbuntemeyer:
Thanks for merging!

Issue #434, opened 2023-05-02
time being recognized by not T?


kthyng:
Is there a way that "time" can be recognized by `cf-xarray` but not "T"? The following works:

```
ds.cf["time"]
```

but not
```
ds.cf["T"]
```

FWIW the time variable looks like this:
<img width="807" alt="image" src="https://user-images.githubusercontent.com/3487237/235795362-9d693a6a-66ca-459a-a4bd-db0da0c7c503.png">

The answer for my questions is always related to `ds.encoding` but I am not sure how to address that here when it looks to me like the results for "time" and "T" should match.
Thanks for any help! cc @dcherian

dcherian:
Yes that should work, this looks like a bug. Can you dig in to it?

kthyng:
Yes but do you have any ideas to start with?

kthyng:
Ok I found the place that causes an error (not sure why this doesn't print out to screen when it happens?):

https://github.com/xarray-contrib/cf-xarray/blob/main/cf_xarray/accessor.py#L324-L326

This line: 
```
results.update((search_in | set(obj.dims)) & sgrid_axes[key])
```

has:
```
sgrid_axes[key]
*** KeyError: 'T'
```

I can dig in more but do you know off the top of your head what is happening in this part?

kthyng:
Ok should I make a new release?

dcherian:
> Ok should I make a new release?

Sure release `0.9.0` after #436 goes in

PR #435, opened 2023-05-03
Don't look for `T` in sgrid axes

kthyng:
cf-xarray is not finding T as a time axis due to a KeyError. This first checks to see if the key is in sgrid_axes before using that.

Closes https://github.com/xarray-contrib/cf-xarray/issues/434

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -321,10 +321,9 @@ def _get_axis_coord(obj: DataArray | Dataset, key: str) -> list[str]:
                     if units in expected:
                         results.update((coord,))
 
-    if key in _AXIS_NAMES and "grid_topology" in obj.cf.cf_roles:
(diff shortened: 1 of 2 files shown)


kthyng:
@dcherian the errors don't look related as far as I can tell.

kthyng:
I'm not sure how other people are using `cf-xarray` currently because this is a total blocker for me!

dcherian:
This looks right, can you add a test.

Presumably you have an sgrid dataset with the `grid_topology` variable, Anyone without it is unaffected

kthyng:
Ok in modifying a test to check for this I was forced to look into `parse_axes` some – I'm not familiar with the grid topology stuff at all. But it looks like X, Y, and Z are the only options anyway for key so I put them directly into `_get_axis_coord`.

Does that work?

dcherian:
Yes I think that's right.

PR #436, opened 2023-05-03
Use pooch instead of bundling standard name table

### LINKED ISSUE #422: Use pooch for the CF standard name table
From @malmans2 in another issue:
> BTW, maybe we could use pooch instead of shipping the table with cf-xarray? I'm 50/50.

I think we should do this. We can provide a download function for backwards compatibility. It'd be useful to hear how people are using that file.

cc @Zeitsperre @aulemahal @kthyng

dcherian:
Closes #422 

cc @Zeitsperre


malmans2 (approved):
Looks good!
I would just change the docstring of argument `source` in `add_canonical_attributes`, to something like:
```
If None, use the latest CF standard name table.
```

malmans2:
Another quick comment, if we use `fsspec` rather than `pooch`, I think it's more likely that it is already installed as it should be a dependency of `dask`.
See: https://github.com/xarray-contrib/cf-xarray/issues/422#issuecomment-1423034942

Zeitsperre:
@dcherian Thanks for the heads up!

dcherian:
>  if we use fsspec rather than pooch, I think it's more likely that it is already installed as it should be a dependency of dask.

Eh, I think its OK. Xarray uses pooch for its tutorial datasets anyway...

PR #437, opened 2023-05-05
use trusted publishing

keewis:
There's a relatively new feature called [trusted publishing](https://docs.pypi.org/trusted-publishers/) on PyPI (based on OpenID Connect) that aims to replace the long-lived API tokens. Github and the action we use to upload to PyPI also [support it](https://github.com/pypa/gh-action-pypi-publish#trusted-publishing).

One main advantage is that this also allows creating *new* PyPI projects, so we don't have to manually upload we a user's API token to create the project, create a restricted API token for the project and use that.

I'm submitting a PR here first before doing the same for `xarray` just to make sure this works just as smoothly for larger projects as for the [little project](https://github.com/umr-lops/xarray-safe-rcm) I just used to try this out.

@dcherian, if this sounds good to you, you (or any other owner of the `cf-xarray` project on PyPI) will have to manually add a pending publisher as described in the PyPI docs linked above, after which the workflow should work as usual.

### DIFF
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -83,6 +83,13 @@ jobs:
     needs: test-built-dist
     if: github.event_name == 'release'
     runs-on: ubuntu-latest
+
+    environment:
+      name: pypi
+      url: https://pypi.org/p/cf-xarray
+    permissions:
+      id-token: write
+
     steps:
       - uses: actions/download-artifact@v3
         with:
@@ -90,6 +97,3 @@ jobs:
           path: dist
       - name: Publish package to PyPI
         uses: pypa/gh-action-pypi-publish@v1.8.5
-        with:
-          password: ${{ secrets.PYPI_PASSWORD }}
-          verbose: true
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi.yaml
```diff
@@ -83,13 +83,17 @@ jobs:
...
+    permissions:
+      id-token: write
+
     steps:
       - uses: actions/download-artifact@v3
         with:
           name: releases
           path: dist
       - name: Publish package to PyPI
         uses: pypa/gh-action-pypi-publish@v1.8.5
```
keewis, line 99:
```suggestion
        uses: pypa/gh-action-pypi-publish@v1.8.5
        with:
          verbose: true
```

dcherian:
Thanks, I think I set it up.

dcherian:
Thanks @keewis That [seemed to work](https://github.com/xarray-contrib/cf-xarray/actions/runs/4932709167/jobs/8816017218) with this warning

```
Notice: Attempting to perform trusted publishing exchange to retrieve a temporary short-lived API token for authentication against https://upload.pypi.org/legacy/ due to __token__ username with no supplied password field
```

keewis:
I think we can just ignore that warning, although it might be worth asking the devs of the PyPI action: `__token__` is the default username, and I've been following the tutorial from their readme.

PR #441, opened 2023-05-19
Add handling of cftime bounds

znichollscr:
At the moment this just adds a failing test that shows that add_bounds fails if the times are cftimes. I'm not sure what the best solution would be.


znichollscr:
@mathause any thoughts on this one? (Calling on a familiar name :))

PR #460, opened 2023-07-17
Add conversion between cf and shapely for lines

jsignell:
I started on this at the SciPy sprints with the idea of pushing forward the work described in this ticket: https://github.com/xarray-contrib/xvec/issues/48. It took me a while to get the logic right :sweat_smile: so I wanted to make sure it was somewhere public in case I don't get a chance to come back to it in the near future.

### DIFF
--- a/cf_xarray/geometry.py
+++ b/cf_xarray/geometry.py
@@ -163,8 +163,10 @@ def cf_to_shapely(ds: xr.Dataset):
     geom_type = ds.geometry_container.attrs["geometry_type"]
     if geom_type == "point":
         geometries = cf_to_points(ds)
-    elif geom_type in ["line", "polygon"]:
+    elif geom_type == "line":
         geometries = cf_to_lines(ds)
+    elif geom_type == "polygon":
+        raise NotImplementedError("Polygon geometry conversion is not implemented.")
     else:
         raise ValueError(
             f"Valid CF geometry types are 'point', 'line' and 'polygon'. Got {geom_type}"
@@ -317,7 +319,8 @@ def cf_to_lines(ds: xr.Dataset):
     # Shorthand for convenience
     geo = ds.geometry_container.attrs
 
-    # The features dimension name, defaults to the one of 'node_count' or the dimension of the coordinates, if present.
+    # The features dimension name, defaults to the one of 'part_node_count'
+    # or the dimension of the coordinates, if present.
     feat_dim = None
     if "coordinates" in geo and feat_dim is None:
         xcoord_name, _ = geo["coordinates"].split(" ")
--- a/cf_xarray/geometry.py
+++ b/cf_xarray/geometry.py
@@ -357,7 +357,7 @@ def cf_to_lines(ds: xr.Dataset):
         if indexes.shape == (0, 2):
             geoms[i] = LineString(xy[j : j + n, :])
         else:
-            geoms[i] = MultiLineString([xy[ii[0]: ii[1], :] for ii in indexes])
+            geoms[i] = MultiLineString([xy[ii[0] : ii[1], :] for ii in indexes])
         j += n
 
     return xr.DataArray(geoms, dims=part_node_count.dims, coords=part_node_count.coords)
--- a/cf_xarray/geometry.py
+++ b/cf_xarray/geometry.py
@@ -111,10 +111,10 @@ def shapely_to_cf(geometries: xr.DataArray | Sequence, grid_mapping: str | None 
     }
     if types.issubset({"Point", "MultiPoint"}):
         ds = points_to_cf(geometries)
-    elif types.issubset({"Polygon", "MultiPolygon"}) or types.issubset(
-        {"LineString", "MultiLineString"}
-    ):
-        raise NotImplementedError("Only point geometries conversion is implemented.")
+    elif types.issubset({"LineString", "MultiLineString"}):
+        ds = lines_to_cf(geometries)
+    elif types.issubset({"Polygon", "MultiPolygon"}):
+        raise NotImplementedError("Polygon geometry conversion is not implemented.")
     else:
         raise ValueError(
             f"Mixed geometry types are not supported in CF-compliant datasets. Got {types}"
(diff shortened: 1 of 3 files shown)


dcherian:
Thanks! This is great to see

cc @aulemahal

jsignell:
Just to make sure this is visible from this side as well: while working on this PR I was pointed to https://github.com/twhiteaker/CFGeom. I have opened [an issue on that repo](https://github.com/twhiteaker/CFGeom/issues/90) to see if the contributors over there have opinions about how these efforts relate to each other.

cf_xarray/tests/test_geometry.py
```diff
@@ -49,6 +49,53 @@ def geometry_ds():
...
+
+    cf_ds = ds.assign(
+        x=xr.DataArray(
+            [0, 1, 4, 5, 0, 1, 1, 1.0, 2.0, 1.7], dims=("node",), attrs={"axis": "X"}
+        ),
+        y=xr.DataArray(
+            [0, 2, 4, 6, 0, 0, 1, 1.0, 2.0, 9.5], dims=("node",), attrs={"axis": "Y"}
+        ),
+        part_node_count=xr.DataArray([4, 3, 3], dims=("index",)),
+        node_count=xr.DataArray([2, 2, 3, 3], dims=("counter",)),
```
jsignell, line 80: This part caught me up for a bit. As I understand it, the `node_count` is the number of nodes per line segment. So each line segment in a multiline results in a value in this array. This means that the dims is not `index` and is not in the shapely version at all. I gave it a kind of random name `counter` which should probably be changed.

cf_xarray/geometry.py
```diff
@@ -293,3 +295,69 @@ def cf_to_points(ds: xr.Dataset):
...
+        part_node_count = ds[part_node_count_name]
+
+    j = 0  # The index of the first node.
+    geoms = np.empty(part_node_count.shape, dtype=object)
+    node_count_cumsum = np.insert(np.cumsum(node_count), 0, 0)
+
+    # i is the feature index, n its number of nodes
+    for i, n in enumerate(part_node_count.values):
+        subset = node_count_cumsum[(j <= node_count_cumsum) & (node_count_cumsum <= n)]
+        indexes = np.stack([subset[:-1], np.roll(subset, 2)[:-1]]).transpose()
```
jsignell, line 356: I was very inspired by a "thinking in arrays" tutorial I went to at SciPy, but if this feels too opaque I can change it to be loopier.

martinfleis: do you think we could actually avoid this loop and use vectorised geometry construction using shapely.linestrings](https://shapely.readthedocs.io/en/stable/reference/shapely.linestrings.html#shapely.linestrings)? (I haven't had a chance to play with your code yet)

jsignell: oooh yeah I can take a look.

jsignell: From my reading it looks like that can only return linestrings and we need to be able to create multilinestrings OR linestrings depending on the number of `node_count` values contained within a given `part_node_count`. I guess we could create a new method on shapely that does this. OR there could be two separate calls instead of this loop. One call to `shapely.linestrings` and one call to `shapely.multilinestrings`

martinfleis: > there could be two separate calls

I think that would be the preferred way. and then sorting according to the original order.

jsignell: ok I gave it a shot. I feel like I am maybe missing some very easy cleanups, but it works, so I figured I'd push.

dcherian:
@jsignell Any interest in picking this back up?

martinfleis:
This could be significantly simplified by using `shapely.to_ragged_arrays` and `from_ragged_arrays`. That is used to create a GeoArrow arrays of coordinates, which are nearly the same as CF representation. Just one uses offsets to indicate where a new geom starts and the others counts of coords per geom, which should be an easy conversion.

jsignell:
The motivation for this work was to support IO in xvec, so I kind of stalled after @martinfleis came up with a better solution on that side. If there is still desire for this work then sure I can pick it back up.

jsignell:
Ah yeah I did mean the WKB option. In that case I can pick this back up. Or @dcherian feel free to take it over if you have some urgency around it.

jsignell:
Thanks for the tip @martinfleis! I looked into `from_ragged_array` and I agree that it is pretty similar. Actually the logic that is in this PR for creating linestrings is nearly identical: https://github.com/shapely/shapely/blob/9540b77a939ba35d066d618183e67fc905ed9ec4/shapely/_ragged_array.py#L334-L337

I think the complexity in this PR comes from how cf intermingles MULTILINESTRING and LINESTRING. So once you have the line segments you still need to figure out which of them are going to be MULTILINESTRING objects. That is basically what the entirety this section is doing: https://github.com/xarray-contrib/cf-xarray/pull/460/files#diff-6484364cf7fa3216828551b2531094b30d77cf393313226a01d425c22580688aR353-R370

That is probably not the best possible way to do that conversion but I don't see anything similar in the `shapely.from_ragged_array` implementation

jsignell:
Ok I added the shapely to CF direction for that one I found it pretty straightforward to use `to_ragged_array` so that one should read pretty cleanly! I'm going to try again now to see if I can simplify the other direction using `from_ragged_array`

jsignell:
Ok this is clearly a case of Monday brain > Friday brain.

dcherian:
I can't really review this  so I'm happy to merge when you're ready.

jsignell:
Yeah it's finally green and fairly well tested so it is actually ready for review for the first time :sweat_smile:

dcherian:
Thanks @jsignell . We can address @martinfleis comments as they come in :)

jsignell:
YOLO!

martinfleis (commented):
Approved (even if too late)! Looks good @jsignell! 

I suppose that Polygons will look similar but we'll need to deal with holes as one layer on top of this.

dcherian:
Thanks for taking a look @martinfleis 🙏

PR #463, opened 2023-08-09
remove distutils

ocefpaf:
It is imported but unused (tests passes without it). Removing this import will make `cf_xarray` ready to Python 3.12.

---
Edit: Changing to packaging.version.parse is not the same thing as `LooseVersion`, see https://discuss.python.org/t/add-looseversion-to-pypa/24359/6,  but we should be OK b/c here b/c the packages checked all follow PEP440.

### DIFF
--- a/cf_xarray/tests/__init__.py
+++ b/cf_xarray/tests/__init__.py
@@ -1,7 +1,6 @@
 import importlib
 import re
 from contextlib import contextmanager
-from distutils import version
 
 import dask
 import pytest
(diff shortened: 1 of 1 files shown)


dcherian:
Cool thanks! I copied that from xarray, so if  you could change it there, that would be great too!

ocefpaf:
> Cool thanks! I copied that from xarray, so if you could change it there, that would be great too!

Will do. The sooner we get rid of distutils the better, or we will see tons of dumb breakages when moving to Python 3.12.

ocefpaf:
Looks like someone already implemented the necessary changes in xarray.

See https://github.com/pydata/xarray/pull/6096

PR #475, opened 2023-10-24
Bump to py3.9, test py3.12

dcherian:


dcherian:
myst-nb only supports py3.10.  I think this is why I was getting weird behaviour?

PR #478, opened 2023-10-26
Add grid_bounds_to_polygons

dcherian:
- [x] Add to api.rst
- [ ] Add test 

This converts "bounds" in the 2-element form to an array of shapely polygons, one per grid cell. A step towards nicely packaging conservative regridding for the ecosystem. It could easily be generalized to the 4-element "vertex" form

Modified from
https://notebooksharing.space/view/c6c1f3a7d0c260724115eaa2bf78f3738b275f7f633c1558639e7bbd75b31456#displayOptions=

cc @martinfleis @benbovy @aulemahal for awareness. I'm not sure if this is at all useful for `xvec` but on the off-chance that it is..

### DIFF
--- a/cf_xarray/geometry.py
+++ b/cf_xarray/geometry.py
@@ -462,7 +462,9 @@ def grid_to_polygons(ds: xr.Dataset) -> xr.DataArray:
 
     grid = ds.cf[["latitude", "longitude"]].load()
     bounds = grid.cf.bounds
+    coordinates = grid.cf.coordinates
     dims = grid.cf.dims
+    bounds_dim = grid.cf.get_bounds_dim_name("latitude")
 
     if "latitude" in dims or "longitude" in dims:
         # for 1D lat, lon, this allows them to be
@@ -471,13 +473,23 @@ def grid_to_polygons(ds: xr.Dataset) -> xr.DataArray:
 
     assert "latitude" in bounds
     assert "longitude" in bounds
+    (lon,) = coordinates["longitude"]
     (lon_bounds,) = bounds["longitude"]
+    (lat,) = coordinates["latitude"]
     (lat_bounds,) = bounds["latitude"]
 
     with xr.set_options(keep_attrs=True):
-        (points,) = xr.broadcast(grid)
+        broadcasted = xr.broadcast(
+            grid[lon],
+            grid[lat],
+            grid[lon_bounds],
+            grid[lat_bounds],
+            exclude=bounds_dim,
+        )
+        asdict = dict(zip([lon, lat, lon_bounds, lat_bounds], broadcasted))
+        # display(asdict)
+        points = xr.Dataset(asdict)
 
-    bounds_dim = grid.cf.get_bounds_dim_name("latitude")
     points = points.transpose(..., bounds_dim)
     lonbnd = points[lon_bounds].data
     latbnd = points[lat_bounds].data
--- a/cf_xarray/geometry.py
+++ b/cf_xarray/geometry.py
@@ -293,3 +293,64 @@ def cf_to_points(ds: xr.Dataset):
         j += n
 
     return xr.DataArray(geoms, dims=node_count.dims, coords=node_count.coords)
+
+
+def grid_bounds_to_polygons(ds: xr.Dataset) -> xr.DataArray:
+    """
+    Converts a regular 2D lat/lon grid to a 2D array of shapely polygons.
+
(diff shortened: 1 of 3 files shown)


cf_xarray/geometry.py
```diff
@@ -293,3 +293,64 @@ def cf_to_points(ds: xr.Dataset):
...
+    # - 4 corners to a polygon, and
+    # - 2 from stacking lat, lon along the last axis
+    # flatten here to make iteration easier. It would be nice to avoid that.
+    # potentially with just np.vectorize. The polygon creation is the real slow bit.
+    # Shapely's MultiPolygon also iterates over a list in Python...
+    blocked = np.stack([expanded_lon, latbnd[..., [0, 1, 1, 0]]], axis=-1).reshape(
+        -1, 4, 2
+    )
+    polyarray = np.array(
+        [Polygon(blocked[i, ...]) for i in range(blocked.shape[0])], dtype="O"
```
dcherian, line 350: This loop is the slow piece. I couldn't see a way to avoid it

cf_xarray/geometry.py
```diff
@@ -293,3 +293,64 @@ def cf_to_points(ds: xr.Dataset):
...
+    )
```
martinfleis, line 351:
```suggestion
    polyarray = shapely.polygons(blocked)
```
This is how to avoid the loop. It gets about 10x faster.

martinfleis: Requires shapely >= 2 but that should optimally be a general requirement now to avoid performance issues.

martinfleis:
In Xvec we could flatten the array and use it as an index instead of latitude and longitude coordinates if there's a use case for that.

cf_xarray/geometry.py
aulemahal: I'm not sure I understand what comes before enough to suggest something complete, but for the curvilinear case I had this in another project:
```
ds = ds.cf.add_bounds(["longitude", "latitude"])
polyarray = shapely.polygons(shapely.linearrings(ds.lon_bounds, ds.lat_bounds))
```

dcherian: Amazing! This is why I love OSS =)

I looked at the docstrings for `Polygon`, `MultiPolygon` but didn't see a reference to `shapely.polygons`. It'd be nice if those functions were more prominent. While there is some user error here (I didn't see "Geometry Creation" under "API reference"), it's not clear at all why those functions are much faster.

dcherian: I'll say both suggestions seem to work. I'll go with @aulemahal's since it avoids the `np.stack`

martinfleis: They are faster because you're shipping the whole array to C and let GEOS deal with the loop, while Polygon creates a scalar geometry. I'll bring the docs feedback to the upstream ;).

cf_xarray/geometry.py
```diff
@@ -438,3 +438,66 @@ def cf_to_lines(ds: xr.Dataset):
...
+        latbnd = latbnd[..., [0, 1, 1, 0]]
+
+    elif points.sizes[bounds_dim] != 4:
+        raise ValueError(
+            f"The size of the detected bounds or vertex dimension {bounds_dim} is not 2 or 4."
+        )
+
+    # geopandas needs this
+    mask = lonbnd[..., 0] >= 180
+    lonbnd[mask, :] = lonbnd[mask, :] - 360
```
mathause, line 496: I think the reason for this is that most geographic data is given in WGS84 which is -180...180 (I may be wrong here). So if you want to compare it to regional polygons it is probably a good idea to wrap the data. Still, it may be confusing to users.

Maybe this could be optional. However, I am not sure what a good name is (In regionmask I call it `wrap_lon` with the options `180` (= what you do here), `360` and `False`. That works but there may be better names).

mathause (commented):
That looks very promising also for regionmask as we discussed in regionmask/regionmask#499. Is it a general assumption in cf_xarray that your grid is lat/ lon in degree?

cf_xarray/geometry.py
```diff
@@ -585,3 +585,66 @@ def cf_to_polygons(ds: xr.Dataset):
     geoms = np.where(np.diff(offset3) == 1, polygons[offset3[:-1]], multipolygons)
 
     return xr.DataArray(geoms, dims=node_count.dims, coords=node_count.coords)
+
+
+def grid_to_polygons(ds: xr.Dataset) -> xr.DataArray:
```
dcherian, line 590:
```suggestion
def bounds_to_polygons(ds: xr.Dataset) -> xr.DataArray:
```

PR #482, opened 2023-12-01
Do not include tests when calculating coverage

jsignell:
Before:
```
---------- coverage: platform linux, python 3.10.13-final-0 ----------
Name                               Stmts   Miss  Cover
------------------------------------------------------
cf_xarray/__init__.py                  8      0   100%
cf_xarray/_version.py                  1      0   100%
cf_xarray/accessor.py               1053     62    94%
cf_xarray/coding.py                   53      7    87%
cf_xarray/criteria.py                 17      2    88%
cf_xarray/datasets.py                 89      0   100%
cf_xarray/formatting.py              144     12    92%
cf_xarray/geometry.py                 98      2    98%
cf_xarray/helpers.py                  64      6    91%
cf_xarray/options.py                  23      0   100%
cf_xarray/scripts/make_doc.py         28      1    96%
cf_xarray/sgrid.py                    20      0   100%
cf_xarray/tests/__init__.py           47     16    66%
cf_xarray/tests/test_accessor.py    1054      2    99%
cf_xarray/tests/test_coding.py        19      0   100%
cf_xarray/tests/test_geometry.py      76      0   100%
cf_xarray/tests/test_helpers.py       38      2    95%
cf_xarray/tests/test_options.py        5      0   100%
cf_xarray/tests/test_scripts.py       13      0   100%
cf_xarray/tests/test_units.py         39      0   100%
cf_xarray/units.py                    41      4    90%
cf_xarray/utils.py                    71      8    89%
------------------------------------------------------
TOTAL                               3001    124    96%
```

After:
```
---------- coverage: platform linux, python 3.10.13-final-0 ----------
Name                            Stmts   Miss  Cover
---------------------------------------------------
cf_xarray/__init__.py               8      0   100%
cf_xarray/_version.py               1      0   100%
cf_xarray/accessor.py            1053     62    94%
cf_xarray/coding.py                53      7    87%
cf_xarray/criteria.py              17      2    88%
cf_xarray/datasets.py              89      0   100%
cf_xarray/formatting.py           144     12    92%
cf_xarray/geometry.py              98      2    98%
cf_xarray/helpers.py               64      6    91%
cf_xarray/options.py               23      0   100%
cf_xarray/scripts/make_doc.py      28      1    96%
cf_xarray/sgrid.py                 20      0   100%
cf_xarray/units.py                 41      4    90%
cf_xarray/utils.py                 71      8    89%
---------------------------------------------------
TOTAL                            1710    104    94%
```

PR #483, opened 2023-12-01
Fix codecov reporting (follow up to #482)

jsignell:
Sorry about that I missed that codecov also needs to be told not to include tests.


jsignell:
This looks more like it.

Issue #491, opened 2024-01-12
flag_masks formatting for larger bit depth


mps01060:
For our project we have ~20 independent flag_masks that can simultaneously be triggered (QA/QC flags). The current implementation of "flag_masks" works well with operations such as "==" and "isin" for integers larger than 8-bit:

```python
import xarray as xr
import numpy as np
import cf_xarray

# Create example 32-bit flag_masks
flag_indep_uint32 = xr.DataArray(
    2**np.arange(32, dtype=np.uint32),
    dims=("time",),
    attrs={
        "flag_masks": [2**i for i in range(32)],
        "flag_meanings": " ".join([f"flag_{i}" for i in range(32)]),
        "standard_name": "flag_independent",
    },
    name="flag_var",
)
```

### Find where flag_0 is True
```python
flag_indep_uint32.cf == "flag_0"
```
```
<xarray.DataArray 'flag_var' (time: 32)>
array([ True, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False])
Dimensions without coordinates: time
```

### Find where flag_0 or flag_1 or flag_31 are True
```python
flag_indep_uint32.cf.isin(["flag_0", "flag_1", "flag_31"])
```
```
<xarray.DataArray 'flag_var' (time: 32)>
array([ True,  True, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False,  True])
Dimensions without coordinates: time
```

The one part that doesn't seem supported is just using the ".cf" to print the flags/bits:
```python
flag_indep_uint32.cf
```
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../python3.10/site-packages/cf_xarray/accessor.py", line 1543, in __repr__
    return ("".join(self._generate_repr(rich=False))).rstrip()
  File ".../lib/python3.10/site-packages/cf_xarray/accessor.py", line 1561, in _generate_repr
    _format_flags(self, rich), title="Flag Variable", rich=rich
  File ".../lib/python3.10/site-packages/cf_xarray/formatting.py", line 200, in _format_flags
    bitstring[abs(b)] = _format_cf_name("1" if b >= 0 else "0", rich)
IndexError: list assignment index out of range
```

[...]

PR #492, opened 2024-01-12
Add larger bit depth for displaying flag_masks

mps01060:
I would like to add larger bit depth support for displaying the flags / flag_masks as in #491 . Thank you for considering this.


dcherian:
Thanks, this LGTM!

PR #493, opened 2024-01-15
Fix conversion between shapely and cf for lines

jsignell:
I started working on conversion between shapely and cf for polygons and in that work I realized that I had misunderstood the roles of `part_node_count` and `node_count` in #460.

My new understanding is mostly based on: "Example 7.16. Polygons with holes" in https://cfconventions.org/cf-conventions/cf-conventions.html#geometries

Even though that example is for polygons it shows clearly that the `part_node_count` is meant to denote the edges of the individual shapes and `node_count` is meant to denote the edges of multi-shapes.


dcherian:
Thanks!

PR #495, opened 2024-01-18
Add conversion between shapely and cf for polygons

jsignell:
This is a follow up to #460 now with polygons!

There is a slight wrinkle in that unlike shapely cf doesn't seem to associate particular holes with particular polygons. I am kind of assuming that the order will be interleaved exterior and interior rings. This seem to match the examples that I've seen.


cf_xarray/geometry.py
```diff
@@ -430,3 +424,163 @@ def cf_to_lines(ds: xr.Dataset):
...
+
+    if coord is not None:
+        ds = ds.assign_coords({dim: coord})
+
+    # Special case when we have no MultiPolygons
+    if len(ds.part_node_count) == len(ds.node_count):
+        ds = ds.drop_vars("part_node_count")
+        del ds.geometry_container.attrs["part_node_count"]
+
+    # Special case when we have not holes
```
aulemahal, line 507:
```suggestion
    # Special case when we have no holes
```

cf_xarray/tests/test_geometry.py
```diff
@@ -126,6 +126,154 @@ def geometry_line_without_multilines_ds():
...
+    from shapely.geometry import Polygon
+
+    # empty/fill workaround to avoid numpy deprecation(warning) due to the array interface of shapely geometries.
+    geoms = np.empty(2, dtype=object)
+    geoms[:] = [
+        Polygon(([50, 0], [40, 15], [30, 0])),
+        Polygon(
+            ([70, 50], [60, 65], [50, 50]),
+            [
+                ([55, 55], [60, 60], [75, 55]),
```
aulemahal, line 179:
```suggestion
                ([55, 55], [60, 60], [65, 55]),
```
I'm guessing a typo. The polygon is otherwise invalid!
Old in red, new in blue.
image

aulemahal (approved):
Nice! I don't have much to say, this looks clean!

The typo in the tests made me think about invalid polygons, but I guess we don't really care at this level ? The invalidity is simply being translated from one format to the other, not our fault.

cf_xarray/tests/test_geometry.py
jsignell: Yes that is definitely a typo. Thanks or the visual! I guess if shapely doesn't complain about invalid shapes neither should we?

cf_xarray/geometry.py
```diff
@@ -430,3 +424,163 @@ def cf_to_lines(ds: xr.Dataset):
...
+            "y": xr.DataArray(y, dims=("node",), attrs={"axis": "Y"}),
+            "crd_x": xr.DataArray(crdX, dims=(dim,), attrs={"nodes": "x"}),
+            "crd_y": xr.DataArray(crdY, dims=(dim,), attrs={"nodes": "y"}),
+        },
+    )
+
+    if coord is not None:
+        ds = ds.assign_coords({dim: coord})
+
+    # Special case when we have no MultiPolygons
```
jsignell, line 502:
```suggestion
    # Special case when we have no MultiPolygons and no holes
```

cf_xarray/tests/test_geometry.py
```diff
@@ -126,6 +126,154 @@ def geometry_line_without_multilines_ds():
...
+            ],
+        ),
+    ]
+
+    ds = xr.Dataset()
+    shp_da = xr.DataArray(geoms, dims=("index",), name="geometry")
+
+    cf_ds = ds.assign(
+        x=xr.DataArray(
+            [50, 40, 30, 50, 70, 60, 50, 70, 55, 60, 75, 55],
```
jsignell, line 189:
```suggestion
            [50, 40, 30, 50, 70, 60, 50, 70, 55, 60, 65, 55],
```

jsignell:
Ok @dcherian I think this one is good to go!

dcherian:
Thanks Julia! Can we have you on board as a maintainer please/

PR #498, opened 2024-01-31
cast units to `str` before applying any other preprocessor

keewis:
The idea is to accept integer units of `1` for dimensionless, which is explicitly allowed by the CF conventions (we could, however, pass through objects that are not integers unmodified).

Note that `pint` itself doesn't really like the idea of integer units, so this will only work with `ureg.parse_units(1)`: `ureg.Quantity(1, 1)` will raise before any preprocessor is called.

- [x] closes an aspect of jbusecke/xMIP#322

### DIFF
--- a/cf_xarray/units.py
+++ b/cf_xarray/units.py
@@ -77,7 +77,12 @@ units = pint.UnitRegistry(
     ],
     force_ndarray_like=True,
 )
+# ----- end block copied from metpy
 
+# need to insert to make sure this is the first preprocessor
+units.preprocessors.insert(0, str)
+
+# -----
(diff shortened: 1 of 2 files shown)


cf_xarray/units.py
```diff
@@ -77,7 +77,12 @@ def repl(m):
     ],
     force_ndarray_like=True,
 )
+# ----- end block copied from metpy
+
+# need to insert to make sure this is the first preprocessor
```
dcherian, line 82:
```suggestion
# need to insert to make sure this is the first preprocessor
# This ensures we convert integer `1` to string `"1"`, as needed by pint.
```

PR #515, opened 2024-06-08
ENH: Update units in .cf.differentiate

DWesl:
Differentiating a field results in a field whose units are the quotient of those of the original field with the units of the field with respect to which the derivative was taken.  This PR adds code to update the units of the derivative to reflect that.

The helper function should simplify adding similar functionality for integrate and cumulative_integrate.

Should I add validation that neither unit is the empty string, since that breaks parsing?

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -953,6 +953,35 @@ def _possible_x_y_plot(obj, key, skip=None):
         return _get_possible(obj.cf, y_criteria)
 
 
+def _update_data_units(
+        result: DataArray | Dataset,
+        source: DataArray | Dataset,
+        coord_name: str,
+        new_unit_template: str
+) -> DataArray | Dataset:
+    try:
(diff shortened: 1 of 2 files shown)


dcherian:
Thanks for the PR.

However, I don't think I will merge this. I'd prefer that `pint` handle these unit transformations instead. Is there a reason `pint` doesn't work for you?

DWesl:
I suppose I could make my own function to wrap the `ds.differentiate()` call and use pint to update the `units` attribute afterwards, but I've gotten used to using the xarray method, and the accessors aren't all that different.

If you'd prefer I move this to pint-xarray rather than cf-xarray, I can do that.

dcherian:
I guess the problem here is that units on indexed coordinates are not supported yet, which would require movement on https://github.com/xarray-contrib/pint-xarray/pull/163#issuecomment-1725341173

Issue #513, opened 2024-05-31
CF accessor adds grid mapping as coordinate


juseg:
Possibly a duplicate of #357 or a discussion topic. Opinions welcomed.

### Description

Since `cf_xarray == 0.8.0` (#391), selecting variables by standard name with `ds.cf[standard_name]` results in a different set of coordinates than selecting them by their short name with `ds[short_name]`. This prevents re-injecting these variables in a compatible (and even the original) dataset using `Dataset.assign()` (and possibly other `xarray` merge operations). Instead trying to `Dataset.assign()` a cf-picked, georeferenced variable into a raises a `MergeError`.

Could you help me to understand if this is a design choice or a bug? In my case it comes as a breaking change and I am not exactly sure how to go about (as a cf-xarray fix or downstream in https://github.com/juseg/hyoga/issues/73). Happy to contribute as I can.

### Miminal example
   
```python
import xarray as xr
import cf_xarray

# make grid-mapped fake data (or use xr.open_dataset)
ds = xr.Dataset({
    'tas': ((), None, {
        'grid_mapping': 'crs',
        'standard_name': 'air_temperature'}),
    'crs': ((), None, {
        'grid_mapping_name': 'latitude_longitude'})})

# ds and ds.cf variables have different coordinates
print('crs' in ds['tas'].coords)  # False
print('crs' in ds.cf['air_temperature'].coords)  # True

# assign a cf-picked variable (only works on cf_xarray < 0.8.0)
ds.assign(new=ds.cf['air_temperature'])  # xarray.core.merge.MergeError
```

### Error message

```python
Traceback (most recent call last):
  File "/home/julien/git/code/hyoga/bugs/cfxr08.py", line 17, in <module>
    ds.assign(new=ds.cf['air_temperature'])  # xarray.core.merge.MergeError
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/xarray/core/dataset.py", line 6080, in assign
    data.update(results)
  File "/usr/lib/python3.12/site-packages/xarray/core/dataset.py", line 4946, in update
    merge_result = dataset_update_method(self, other)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/xarray/core/merge.py", line 1104, in dataset_update_method
    return merge_core(
           ^^^^^^^^^^^
[...]

PR #516, opened 2024-06-11
Assign new variables by standard names

juseg:
This pull request implements `ds.cf.assign(**standard_variables)`. The new method should:

- [ ] Assign variables by standard name as `**kwargs` or dictionary.
- [ ] Decide on the variable short names without user input (see below).
- [ ] Either handle or reject duplicate standard names (see below).
- [ ] Workaround xarray incompatibilities, currently #513.

Also needed:

- [ ] Add documentation, where should it go?
- [ ] Add tests, including name duplicates.

I would welcome opinions on the two points below.

## 1. Decide about the variable short name

I propose this algorithm:

- If variable has a short name, absent from the dataset, use it.

```python
ds.cf.assign(air_temperature=DataArray(name='tas')) -> 'tas'
```

- Otherwise, use the standard name as a short variable name.

```python
ds.cf.assign(air_temperature=DataArray()) -> 'air_temperature'
```

- If name is already used in dataset, warn user, and add trailing underscores.

```python
ds.cf.assign(
   air_temperature=DataArray(name=tas)).cf.assign(
   total_precipitation=DataArray(name=tas)) -> tas, tas_
```

- If name is taken by another variable in call, also add trailing underscores.

```python
ds.cf.assign(
   air_temperature=DataArray(name=tas),
   total_precipitation=DataArray(name=tas)) -> tas, tas_
```

## 2. When dataset already contains standard name

```python
ds = xr.Dataset()
ds = ds.cf.assign(air_temperature=0)
ds = ds.cf.assign(air_temperature=1)
```

I find it more difficult to decide what the method should do here.

- Override existing variable (same as `Dataset.assign`).
- Override but raise a warning.
- Assign new variable with same standard name.
- Assign new variable but raise a warning.
- Customize via keyword e.g. `existing: ignore, override, raise, warn`

Note: if we allow multiple variables with the same standard name, the resulting Dataset is technically valid, and `ds.cf` shows several variables associated with one standard name, while `ds.cf[standard_name] fails with a `KeyError`.

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2633,6 +2633,54 @@ class CFDatasetAccessor(CFAccessor):
                 results[v].append(k)
         return results
 
+    def assign(self, **standard_variables: Any) -> Dataset:
+        """Assign new variables by standard_name.
+
+        Parameters
+        ----------
+        standard_variables : mapping
+            A mapping from netCDF Climate and Forecast compliant standard names
+            (http://cfconventions.org/standard-names.html), to the new values.
+            The new values are passed to ``:meth:Dataset.assign``.
+
+        Returns
+        -------
+        dataset : Dataset
+            A new dataset with the new variables added, replacing any existing
+            variable with the same standard name. New variables are labelled
+            according to their input (short) names, or standard names if short
+            names are missing, with added trailing underscores in the event of
+            duplicates.
+
+        See Also
+        --------
+        DataSet.assign
+        """
+
+        # initalize dictionary mapping short names to the new variables
+        variables = {}
+
+        # for each standard name and value pair
+        for standard_name, values in standard_variables.items():
+
+            # default to using existing short name or standard name
+            name = getattr(values, "name", standard_name)
+
+            # add trailing underscores until we find a free slot
+            while name in self._obj.data_vars or name in variables:
+                emit_user_level_warning(
+                    f"found existing variable {name}, using {name}_ instead",
+                    UserWarning)
+                name += '_'
+
+            # map name to values (without coords, see #513, xarray#6447)
+            values = xr.as_variable(values)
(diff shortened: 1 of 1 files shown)


dcherian:
I think the cf-xarray version of this only really makes sense when the assigned name is a standard name on one of the present variables, so (2) in your listing.

For (1), we should just forward on to Xarray, as usual.

>  if we allow multiple variables with the same standard name, the resulting Dataset is technically valid, and ds.cf shows several variables associated with one standard name, while ds.cf[standard_name] fails with a KeyError`

Yes, this is intentional. `ds.cf[standard_name]` will raise an error unless there is only one result, since that is the only way to return a DataArray. to get all, use `ds.cf[[standard_name]]`. Then you will get a dataset with all dataarrays with that standard name.

cf_xarray/accessor.py
```diff
@@ -2633,6 +2633,54 @@ def grid_mapping_names(self) -> dict[str, list[str]]:
...
+        DataSet.assign
+        """
+
+        # initalize dictionary mapping short names to the new variables
+        variables = {}
+
+        # for each standard name and value pair
+        for standard_name, values in standard_variables.items():
+            # default to using existing short name or standard name
+            name = getattr(values, "name", standard_name)
```
dcherian, line 2666: Something like 
https://github.com/xarray-contrib/cf-xarray/blob/52160a51c66edbec84f1412890887a7874dc2d25/cf_xarray/accessor.py#L2473

should be used here. So

```
mapped_name, = apply_mapper(_get_all, self._obj, standard_name, error=False, default=[standard_name])
```
 will do the translation to actual name in the dataset, if possible. Use this to create a new dictionary that you can then pass to `self._obj.assign`

juseg: Thank you for the feedback, I will be looking at this next week.

PR #517, opened 2024-06-21
Add geometry encoding and decoding functions.

dcherian:
These differ from `shapely_to_cf` and `cf_to_shapely` by returning all variables. Those function, simply encode and decode geometry-related variables.

cc @jsignell 

xref https://github.com/xarray-contrib/xvec/issues/26

### DIFF
--- a/cf_xarray/geometry.py
+++ b/cf_xarray/geometry.py
@@ -159,6 +159,8 @@ def encode_geometries(ds: xr.Dataset):
     for varname, var in ds._variables.items():
         if varname == name:
             continue
+        # TODO: this is incomplete. It works for vector data cubes where one of the geometry vars
+        # is a dimension coordinate.
         if name in var.dims:
             var = var.copy()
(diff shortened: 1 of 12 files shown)


cf_xarray/geometry.py
```diff
@@ -1,11 +1,188 @@
...
+    """
+    Encode any discovered geometry variables using the CF conventions.
+
+    Practically speaking, geometry variables are numpy object arrays where the first
+    element is a shapely geometry.
+
+    .. warning::
+
+       Only a single geometry variable is supported at present. Contributions to fix this
+       are welcome.
```
martinfleis, line 94: I think that is actually a limitation of CF conventions, no? I have a vague memory that it supports at max one geometry variable.

dcherian: Each variable is only linked to one geometry variable through the `geometry` attribute. But I can't see why you can't have multiple geometry variables.

PR #520, opened 2024-06-25
Cache flag-related objects and refactor extraction of flags

Descanonge:
The dictionary of flag values/masks is cached in the CFAccessor, moving the code of `create_flag_dict` into a property of the accessor.

The dataset of booleans masks is cached in the CFDataArrayAccessor. I don't know if this is a welcomed change. I would argue that if it is presented as a property (`.flags`), I would expect it to be the same instance on different calls, contrary to a method like `.flags()` or `.get_flags()`. But I'm not very familiar with the rest of the cf-xarray API. Anyway it can be reverted easily.

I have also refactored the construction of that flag dataset to make it (hopefully) more readable and understable. It makes use of the FlagParams named-tuple. The operations on the data are unchanged.

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2832,6 +2832,11 @@ class CFDatasetAccessor(CFAccessor):
 
 @xr.register_dataarray_accessor("cf")
 class CFDataArrayAccessor(CFAccessor):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._flags: Dataset | None = None
+
     @property
     def formula_terms(self) -> dict[str, str]:  # numpydoc ignore=SS06
         """
@@ -2977,6 +2982,8 @@ class CFDataArrayAccessor(CFAccessor):
         """
         Dataset containing boolean masks of available flags.
         """
+        if self._flags is not None:
+            return self._flags
         return self._extract_flags()
 
     def _extract_flags(self, flags: Sequence[Hashable] | None = None) -> Dataset:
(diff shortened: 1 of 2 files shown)


cf_xarray/accessor.py
```diff
@@ -1157,23 +1122,64 @@ class CFAccessor:
...
+        Return possible flag meanings and associated bitmask/values.
+
+        The mapping values are a tuple containing a bitmask and a value. Either
+        can be None.
+        If only a bitmask: Independent flags.
+        If only a value: Mutually exclusive flags.
+        If both: Mix of independent and mutually exclusive flags.
+        """
+        if self._flag_dict is not None:
+            return self._flag_dict
```
dcherian, line 1142: The problem is that you can change the underlying values in the DataArray and then the flags will be wrong.

dcherian:
> I have also refactored the construction of that flag dataset to make it (hopefully) more readable and understable.

Nice!

Re: API if it is useful to get `flag_dict` then perhaps we should create a `Flags` dataclass with `as_dict` and `as_dataset` methods?

cf_xarray/accessor.py
Descanonge: I forgot that if for DataArray.data most operations are not done in place (you get a new DataArray), changing `DataArray.attrs` is easily done in place. I only see either checking if any of the attributes have changed, or urging to use [`DataArray.assign_attrs()`](https://docs.xarray.dev/en/latest/generated/xarray.DataArray.assign_attrs.html) in the documentation...

PR #523, opened 2024-06-28
Rewrite unit formatter for pint 0.24 and earlier

### LINKED ISSUE #522: CF unit formatter incompatible with latest pint
With pint 0.24, the formatting machinery changed a lot. In 0.24.0 our formatter broke because we couldn't re-create Unit object, the registry was None (see hgrecco/pint#2009), this was fixed, but another bugfix broke the specific case of a dimensionless variable, see hgrecco/pint#2024.

MWE with cf-xarray
```python3
from cf_xarray import units

u = units.units.parse_units("")
f"{u:cf}"
```
raises the error shown in the pint issue.

aulemahal:
Fixes #522 and hgrecco/pint#2024.

- Rewrite of the custom unit formatter to properly use the tools of pint instead of relying on regex.
- To preserve backwards compatibility the `pint.formatter` call  changes according to the installed version of  `pint`.
- Before, `f"{u:cf}"` and `f"{u:~cf}"` returned the same thing except for "dimensionless". This shortening is done explicitly in the new formatter with a `if-else` in case it was already done by `pint`.
- Removed a few unneeded lines 
  +  "percent" was always printed as "%" for all supported pint versions, so I don't know why it as there.
  + The formatter's declaration never raised ImportError, so I think the `try-except` was a legacy from something else. 
- Added some warnings to the documentation about non-udunits units.
- Added dimensionless to the tests.

I tested this with pint 0.24.1 and 0.23. I also did a very minimal test with pint 0.22 down to 0.12. This seems to support pint >= 0.18, like the previous version did.

With pint < 0.24 the only difference this makes is that dimensionless units are returned as `""` with `{u:cf}` and `{u:~cf}`, where as the previous version returned `"dimensionless"` for `{u:cf}`.

### DIFF
--- a/cf_xarray/units.py
+++ b/cf_xarray/units.py
@@ -27,9 +27,10 @@ def short_formatter(unit, registry, **options):
     out : str
         Units following CF-Convention, using symbols.
     """
-    # pint 0.24.1 gives this for dimensionless units
-    if unit == {"dimensionless": 1}:
-        return ""
+    # pint 0.24.1 gives {"dimensionless": 1} for non-shortened dimensionless units
+    # CF uses "1" to denote fractions and dimensionless quantities
+    if unit == {"dimensionless": 1} or not unit:
+        return "1"
 
     # If u is a name, get its symbol (same as pint's "~" pre-formatter)
     # otherwise, assume a symbol (pint should have already raised on invalid units before this)
@@ -51,7 +52,7 @@ def short_formatter(unit, registry, **options):
         )
 
     out = pint.formatter(*args, as_ratio=False, product_fmt=" ", power_fmt="{}{}")
-    # To avoid potentiel unicode problems in netCDF. In both case, this unit is not recognized by udunits
+    # To avoid potentiel unicode problems in netCDF. In both cases, this unit is not recognized by udunits
     return out.replace("Δ°", "delta_deg")
 
 
--- a/cf_xarray/units.py
+++ b/cf_xarray/units.py
@@ -4,62 +4,56 @@ import functools
 import re
 
 import pint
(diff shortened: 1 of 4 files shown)


dcherian:
Failure is real.

keewis:
I've been a bit away from `pint`, so I'll have to first look into how it changed in the most recent version before giving any thorough reviews, but at a quick glance this looks fine.

dcherian:
> but at a quick glance this looks fine.

Thanks! no need for a thorough review.

cf_xarray/units.py
```diff
@@ -4,62 +4,56 @@
...
+        Additional options (may be ignored)
+
+    Returns
+    -------
+    out : str
+        Units following CF-Convention, using symbols.
+    """
+    # pint 0.24.1 gives this for dimensionless units
+    if unit == {"dimensionless": 1}:
+        return ""
```
MuellerSeb, line 32: Why not `return "dimensionless"` like before? I know pint 0.24.1 is doing this now, but since this is a custom formatter, this could be used to keep it backward compatible, or?

aulemahal: "dimensionless" is not recognized by udunits, so we shouldn't return it. Prior to this PR, `"{u:cf}" flagged (in the doc) as returning an invalid unit because of this (ok it may not have been so clear). This PR fixes this issue.

However, returning to the conventions, I see that the empty string is indeed not really recognized either. The canonical string for dimensionless unit should maybe be "1" instead ? With the new pint we can do that.

http://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#dimensionless-units

PR #526, opened 2024-07-01
Allow encoding/decoding multiple geometries

dcherian:
This needs
- updated docstrings
- adding GeometryNames to api.rst
- loot more tests
- typing fixes

but it works with all xvec examples!

### DIFF
--- a/cf_xarray/geometry.py
+++ b/cf_xarray/geometry.py
@@ -10,6 +10,7 @@ import pandas as pd
 import xarray as xr
 
 GEOMETRY_CONTAINER_NAME = "geometry_container"
+FEATURES_DIM_NAME = "features"
 
 __all__ = [
     "decode_geometries",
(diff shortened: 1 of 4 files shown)


cf_xarray/geometry.py
```diff
@@ -30,6 +33,122 @@
...
+
+    @property
+    def geometry_container_attrs(self) -> dict[str, str]:
+        return {
+            "node_count": self.node_count,
+            "node_coordinates": f"{self.node_coordinates_x} {self.node_coordinates_y}",
+            "coordinates": f"{self.coordinates_x} {self.coordinates_y}",
+        }
+
+    def coords(self, *, dim: str, x, y, crdX, crdY) -> dict[str, xr.DataArray]:
```
dcherian, line 91: TODO: save `dim` as `self.core_dim`? This is the core dimension for the decoded shapely geometries.

cf_xarray/geometry.py
```diff
@@ -410,11 +533,12 @@ def points_to_cf(pts: xr.DataArray | Sequence):
     from shapely.geometry import MultiPoint
 
     if isinstance(pts, xr.DataArray):
+        # TODO: Fix this hardcoding
```
dcherian, line 536: `assert pts.ndim == 1`

dcherian:
Could use more tests but I'm unlikely to get to it soon

PR #528, opened 2024-07-18
Adds support for parametric vertical coordinate

jasonb5:
Adds support for parametric vertical coordinates from [CFConventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#parametric-v-coord).

- [ ] Update [docs](https://cf-xarray.readthedocs.io/en/latest/parametricz.html)
- [x] Add name to CITATION.cff
- [x] Remove `squeeze`

### DIFF
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2705,7 +2705,8 @@ class CFDatasetAccessor(CFAccessor):
                         f"Variable {value!r} is required to decode coordinate for {dim!r}"
                         " but it is absent in the Dataset."
                     )
-                terms[key] = ds[value]
+                # keys should be case insensitive
+                terms[key.lower()] = ds[value]
 
             try:
                 func = parametric.func_from_stdname(stdname)
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2707,12 +2707,16 @@ class CFDatasetAccessor(CFAccessor):
                     )
                 terms[key] = ds[value]
 
-            func = parametric.get_parametric_func(stdname)
+            try:
+                func = parametric.func_from_stdname(stdname)
+            except AttributeError:
+                # Should occur since stdname is check before
+                raise NotImplementedError(
+                    f"Coordinate function for {stdname!r} not implmented yet. Contributions welcome!"
+                ) from None
 
-            absent_terms = func._requirements - set(terms)
-
-            if absent_terms:
-                raise KeyError(f"Required terms {absent_terms} absent in dataset.")
+            # let KeyError propagate
+            parametric.check_requirements(func, terms)
 
             ds.coords[zname] = func(**terms)
 
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -25,7 +25,7 @@ from xarray.core.resample import Resample
 from xarray.core.rolling import Coarsen, Rolling
 from xarray.core.weighted import Weighted
 
-from . import sgrid
+from . import parametric, sgrid
 from .criteria import (
     _DSG_ROLES,
     cf_role_criteria,
@@ -2668,13 +2668,8 @@ class CFDatasetAccessor(CFAccessor):
         """
         ds = self._obj
 
-        requirements = {
-            "ocean_s_coordinate_g1": {"depth_c", "depth", "s", "C", "eta"},
-            "ocean_s_coordinate_g2": {"depth_c", "depth", "s", "C", "eta"},
-            "ocean_sigma_coordinate": {"sigma", "eta", "depth"},
-        }
-
         allterms = self.formula_terms
+
         for dim in allterms:
             if prefix is None:
                 assert (
@@ -2696,6 +2691,7 @@ class CFDatasetAccessor(CFAccessor):
                 suffix = dim.split("_")
                 zname = f"{prefix}_" + "_".join(suffix[1:])
 
+            # never touched, if standard name is missing it's not included in allterms
             if "standard_name" not in ds[dim].attrs:
                 continue
             stdname = ds[dim].attrs["standard_name"]
@@ -2704,46 +2700,21 @@ class CFDatasetAccessor(CFAccessor):
             terms = {}
             for key, value in allterms[dim].items():
                 if value not in ds:
+                    # is this ever hit, if variable is missing it's missing in decoded allterms
                     raise KeyError(
                         f"Variable {value!r} is required to decode coordinate for {dim!r}"
                         " but it is absent in the Dataset."
                     )
                 terms[key] = ds[value]
 
-            absent_terms = requirements[stdname] - set(terms)
+            func = parametric.get_parametric_func(stdname)
+
+            absent_terms = func._requirements - set(terms)
+
             if absent_terms:
                 raise KeyError(f"Required terms {absent_terms} absent in dataset.")
 
-            if stdname == "ocean_s_coordinate_g1":
-                # S(k,j,i) = depth_c * s(k) + (depth(j,i) - depth_c) * C(k)
-                S = (
-                    terms["depth_c"] * terms["s"]
-                    + (terms["depth"] - terms["depth_c"]) * terms["C"]
-                )
-
-                # z(n,k,j,i) = S(k,j,i) + eta(n,j,i) * (1 + S(k,j,i) / depth(j,i))
-                ztemp = S + terms["eta"] * (1 + S / terms["depth"])
-
-            elif stdname == "ocean_s_coordinate_g2":
-                # make sure all necessary terms are present in terms
-                # (depth_c * s(k) + depth(j,i) * C(k)) / (depth_c + depth(j,i))
-                S = (terms["depth_c"] * terms["s"] + terms["depth"] * terms["C"]) / (
-                    terms["depth_c"] + terms["depth"]
-                )
-
-                # z(n,k,j,i) = eta(n,j,i) + (eta(n,j,i) + depth(j,i)) * S(k,j,i)
-                ztemp = terms["eta"] + (terms["eta"] + terms["depth"]) * S
-
-            elif stdname == "ocean_sigma_coordinate":
-                # z(n,k,j,i) = eta(n,j,i) + sigma(k)*(depth(j,i)+eta(n,j,i))
-                ztemp = terms["eta"] + terms["sigma"] * (terms["depth"] + terms["eta"])
-
-            else:
-                raise NotImplementedError(
-                    f"Coordinate function for {stdname!r} not implemented yet. Contributions welcome!"
-                )
-
-            ds.coords[zname] = ztemp
+            ds.coords[zname] = func(**terms)
 
 
 @xr.register_dataarray_accessor("cf")
--- a/cf_xarray/accessor.py
+++ b/cf_xarray/accessor.py
@@ -2789,17 +2789,14 @@ class CFDatasetAccessor(CFAccessor):
                 terms[key.lower()] = ds[value]
 
             try:
-                func = parametric.func_from_stdname(stdname)
-            except AttributeError:
+                transform = parametric.TRANSFORM_FROM_STDNAME[stdname]
+            except KeyError:
                 # Should occur since stdname is check before
                 raise NotImplementedError(
                     f"Coordinate function for {stdname!r} not implmented yet. Contributions welcome!"
                 ) from None
 
-            # let KeyError propagate
-            parametric.check_requirements(func, terms)
-
-            ds.coords[zname] = func(**terms)
+            ds.coords[zname] = transform.from_terms(terms)
 
 
 @xr.register_dataarray_accessor("cf")
--- a/cf_xarray/parametric.py
+++ b/cf_xarray/parametric.py
@@ -29,7 +29,7 @@ OCEAN_STDNAME_MAP = {
 }
 
 
-def _derive_ocean_stdname(**kwargs):
+def _derive_ocean_stdname(*, zlev=None, eta=None, depth=None):
     """Derive standard name for computer ocean coordinates.
 
     Uses the concatentation of formula terms `zlev`, `eta`, and `depth`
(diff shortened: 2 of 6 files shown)


cf_xarray/accessor.py
```diff
@@ -2784,46 +2780,26 @@ def decode_vertical_coords(self, *, outnames=None, prefix=None):
...
-                # z(n,k,j,i) = eta(n,j,i) + (eta(n,j,i) + depth(j,i)) * S(k,j,i)
-                ztemp = terms["eta"] + (terms["eta"] + terms["depth"]) * S
-
-            elif stdname == "ocean_sigma_coordinate":
-                # z(n,k,j,i) = eta(n,j,i) + sigma(k)*(depth(j,i)+eta(n,j,i))
-                ztemp = terms["eta"] + terms["sigma"] * (terms["depth"] + terms["eta"])
-
-            else:
+            try:
+                func = parametric.func_from_stdname(stdname)
```
dcherian, line 2792:
```suggestion
                transform = parametric.transform_from_stdname(stdname)
```

nit: I prefer the more descriptive "transform" over "func"

cf_xarray/parametric.py
```diff
@@ -0,0 +1,691 @@
...
+        Name of the function.
+
+    Raises
+    ------
+    AttributeError
+        If a function name `stdname` is not in the module.
+    """
+    m = sys.modules[__name__]
+
+    return getattr(m, stdname)
```
dcherian, line 187: Let's just use a hardcoded dictionary instead"

```
TRANSFORM_FROM_STDNAME = {
"atmosphere_ln_pressure_coordinate": atmosphere_ln_pressure_coordinate,
...
}
```

cf_xarray/parametric.py
```diff
@@ -0,0 +1,691 @@
...
+
+ocean_stdname_map = {
+    "altitude": {
+        "zlev": "altitude",
+        "eta": "sea_surface_height_above_geoid",
+        "depth": "sea_floor_depth_below_geoid",
+    },
+    "height_above_geopotential_datum": {
+        "zlev": "height_above_geopotential_datum",
+        "eta": "sea_surface_height_above_ geopotential_datum",
```
dcherian, line 15:
```suggestion
        "eta": "sea_surface_height_above_geopotential_datum",
```

many of these errors below

cf_xarray/parametric.py
```diff
@@ -0,0 +1,691 @@
...
+    ------
+    AttributeError
+        If a function name `stdname` is not in the module.
+    """
+    m = sys.modules[__name__]
+
+    return getattr(m, stdname)
+
+
+def derive_dimension_order(output_order, **dim_map):
```
dcherian, line 190: This is quite clever, but I'd prefer if the user just did the transpose on their own after this decoding.

cf_xarray/parametric.py
```diff
@@ -0,0 +1,691 @@
...
+        )
+
+        raise ValueError(
+            f"Could not derive standard name from combination of {stdnames}."
+        )
+
+    return found_stdname
+
+
+def check_requirements(func, terms):
```
dcherian, line 113: This is also quite clever. But I'd think it'd be more readable if we instead wrote a dataclass per transform

```python

@dataclass
class AtmosphereLnPressureCoordinate():
    p0: int | float
    lev: xr.DataArray

    def __init__(self, *, p0, lev):
        self.p0 = p0
        self.lev = lev

    @classmethod
    def from_terms(cls, terms: dict):
        # verify terms here.
        # set properties
        return cls(p0=p0, lev=lev)

    def decode(self):
        # do calculation
        pass
```

cf_xarray/parametric.py
```diff
@@ -0,0 +1,691 @@
...
+        A DataArray with new pressure coordinate.
+
+    References
+    ----------
+    Please refer to the CF conventions document :
+      1. https://cfconventions.org/cf-conventions/cf-conventions.html#atmosphere-natural-log-pressure-coordinate
+    """
+    p = p0 * np.exp(-lev)
+
+    p = p.squeeze().rename("p").assign_attrs(standard_name="air_pressure")
```
dcherian, line 256:
```suggestion
    p = p.squeeze().assign_attrs(standard_name="air_pressure")
```

I don't think we need to hardcode these names since we set them when assigning to the dataset in `.cf.decode_vertical_coords`

dcherian:
Thanks for taking this on! I've left a few high-level comments. Mostly I think it would be a lot clearer if we define a dataclass per transform that took care of the all the specialized logic.

cf_xarray/parametric.py
dcherian: Yes with Xarray we care much less about dimension order.

cf_xarray/parametric.py
dcherian: Now that i look at this, it would be good to define an ABC that all the individual coordinates subclass from, something like

```python
from abc import ABC, abstractmethod

class ParametricVerticalCoordinate(ABC):
    @abstractmethod
    @classmethod
    def from_terms(cls, terms: dict):
        # verify terms here.
        ...

    @abstractmethod
    def decode(self):
        ...

    @property
    @abstractmethod
    def computed_standard_name(self):
        ...
```

cf_xarray/parametric.py
pochedls: @jasonb5 / @dcherian – thanks for all of your work on this. I wanted to weigh-in as a user. It would be unexpected to get a matrix of `[time, lat, lon, *plev*]` instead of `[time, *plev*, lat, lon]`. I think users would prefer the former (matches most climate data, CF conventions, and almost certainly the order of the original data, which is typically `[time, *lev*, lat lon]`. 

If this doesn't match cf-xarray's style, xcdat could apply this functionality after calling cf-xarray's transform.

dcherian: Yes xcdat could do `.cf.transpose('T', 'Z', 'Y', 'X')`

> almost certainly the order of the original data, which is typically [time, *lev*, lat lon]).

Is this ordering not preserved here?

pochedls: I just did a quick check: it looks like it goes from (`ds.ta.dims`):

`('time', 'lev', 'lat', 'lon')`

to (`p = ds.a * ds.p0 + ds.b * ds.ps; p.dims`):

`('lev', 'time', 'lat', 'lon')`

In other words, the dataarray of interest is `[time, lev, lat, lon]`, but the transform will put `lev` at the front `[lev, time, lat, lon]`

cf_xarray/parametric.py
```diff
@@ -0,0 +1,864 @@
...
+        xr.DataArray
+            Decoded parametric vertical coordinate.
+        """
+        z_shape = list(self.eta.shape)
+
+        z_shape.insert(1, self.sigma.shape[0])
+
+        z_dims = list(self.eta.dims)
+
+        z_dims.insert(1, self.sigma.dims[0])
```
dcherian, line 719:
```suggestion
        z_shape = (self.sigma.shape[0], *self.eta.shape)
        z_dims = (self.sigma.dims[0], *self.eta.dims)
```

dcherian:
ignore the 3.9 failures, I'll drop it soon.

cf_xarray/parametric.py
```diff
@@ -0,0 +1,864 @@
+from abc import ABC, abstractmethod
+from collections.abc import Sequence
+
+import numpy as np
+import xarray as xr
+from xarray import DataArray
+
+ocean_stdname_map = {
```
dcherian, line 8:
```suggestion
OCEAN_STDNAME_MAP = {
```

Python convention is to make global variables all caps.

cf_xarray/parametric.py
```diff
@@ -0,0 +1,864 @@
...
+        If `kwargs` is empty, missing values for `kwargs` keys, or could not derive the standard name.
+
+    References
+    ----------
+    Please refer to the CF conventions document :
+      1. https://cfconventions.org/cf-conventions/cf-conventions.html#table-computed-standard-names
+    """
+    found_stdname = None
+
+    allowed_names = {"zlev", "eta", "depth"}
```
dcherian, line 67:
```suggestion
```

cf_xarray/parametric.py
```diff
@@ -0,0 +1,864 @@
...
+    },
+    "height_above_mean_sea_level": {
+        "zlev": "height_above_mean_sea_level",
+        "eta": "sea_surface_height_above_mean_sea_level",
+        "depth": "sea_floor_depth_below_mean_sea_level",
+    },
+}
+
+
+def _derive_ocean_stdname(**kwargs):
```
dcherian, line 32:
```suggestion
def _derive_ocean_stdname(*, zlev=None, depth=None, eta=None) -> ???
```

This way the language will help you ensure correctness :)

cf_xarray/parametric.py
```diff
@@ -0,0 +1,864 @@
...
+
+    References
+    ----------
+    Please refer to the CF conventions document :
+      1. https://cfconventions.org/cf-conventions/cf-conventions.html#atmosphere-natural-log-pressure-coordinate
+    """
+
+    def __init__(self, p0, lev):
+        self.p0 = p0
+        self.lev = lev
```
dcherian, line 150:
```suggestion
@dataclass
class AtmosphereLnPressure(ParamerticVerticalCoordinate):
    """Atmosphere natural log pressure coordinate.

    Standard name: atmosphere_ln_pressure_coordinate

    Parameters
    ----------
    p0 : xr.DataArray
        Reference pressure.
    lev : xr.DataArray
        Vertical dimensionless coordinate.

    References
    ----------
    Please refer to the CF conventions document :
      1. https://cfconventions.org/cf-conventions/cf-conventions.html#atmosphere-natural-log-pressure-coordinate
    """

	p0: ???
	lev: ???
```

and so on for the rest please.

cf_xarray/parametric.py
```diff
@@ -0,0 +1,864 @@
...
+
+    @property
+    def computed_standard_name(self) -> str:
+        """Computes coordinate standard name."""
+        return "air_pressure"
+
+    @classmethod
+    def from_terms(cls, terms: dict):
+        """Create coordinate from terms."""
+        b, ps, p0, a, ap = get_terms(terms, "b", "ps", optional=("p0", "a", "ap"))
```
dcherian, line 279: if `get_terms` returned a dict with the right names then we could do 

```
cls(**get_terms(...))
```

and put all the error checking in `__post_init__` (once you move to the dataclass.

I think I prefer that.

dcherian:
Thanks! This is getting close. Just some style nits. 

I haven't looked at the tests but I'm assuming you have it covered with some real life usage too? :)

dcherian:
Please also add your name to CITATION.cff

cf_xarray/parametric.py
```diff
@@ -0,0 +1,815 @@
...
+
+    @property
+    def computed_standard_name(self) -> str:
+        """Computes coordinate standard name."""
+        orog_stdname = self.orog.attrs["standard_name"]
+
+        if orog_stdname == "surface_altitude":
+            out_stdname = "altitude"
+        elif orog_stdname == "surface_height_above_geopotential_datum":
+            out_stdname = "height_above_geopotential_datum"
```
dcherian, line 340:
```suggestion
            out_stdname = "height_above_geopotential_datum"
        else:
            raise ValueError(f"Unknown standard name for hybrid height coordinate: {orog_stdname!r}")
```

cf_xarray/parametric.py
```diff
@@ -0,0 +1,815 @@
...
+    def computed_standard_name(self) -> str:
+        """Computes coordinate standard name."""
+        ztop_stdname = self.ztop.attrs["standard_name"]
+
+        if ztop_stdname == "altitude_at_top_of_atmosphere_model":
+            out_stdname = "altitude"
+        elif (
+            ztop_stdname == "height_above_geopotential_datum_at_top_of_atmosphere_model"
+        ):
+            out_stdname = "height_above_geopotential_datum"
```
dcherian, line 406:
```suggestion
            out_stdname = "height_above_geopotential_datum"
        else:
            raise ValueError(f"Unknown standard name: {out_stdname!r}")
```

cf_xarray/parametric.py
```diff
@@ -0,0 +1,815 @@
...
+            2 * self.a / (self.z1 - self.z2) * (self.depth - self.href)
+        )
+
+        z = xr.where(
+            self.sigma.k <= self.k_c,
+            self.sigma * f,
+            f + (self.sigma - 1) * (self.depth - f),
+        )
+
+        return z.squeeze().assign_attrs(standard_name=self.computed_standard_name)
```
dcherian, line 766:
```suggestion
        return z.assign_attrs(standard_name=self.computed_standard_name)
```

Can we delete all the `squeeze` too please

jasonb5: @dcherian Just want to check, if I remove the squeeze we'll end up with placeholder dimensions e.g. `(lev: 3, dim_0: 1, lat: 2, lon: 2)`. If that's fine then I'll go ahead and remove the call.

dcherian: where is the `dim_0` being created?

cf_xarray/tests/test_parametric.py
```diff
@@ -0,0 +1,613 @@
...
+        name="z",
+    )
+
+    assert_allclose(output, expected)
+
+    assert output.attrs["standard_name"] == "altitude"
+
+
+def test_ocean_double_sigma_coordinate():
+    k_c = xr.DataArray(
```
dcherian, line 502: shouldn't these be scalars? How are they stored in the netCDF files you're looking at?

cf_xarray/parametric.py
jasonb5: @dcherian Circling back around, I've been on leave and just got a chance to look into this. It appears the `dim_0` was being generated by the dimensionless constants. It ends up `xr.DataArray([10], name='p0')` needed to be `xr.DataArray(10, name='p0')`. I've removed the squeeze calls and fixed the constant DataArrays.

cf_xarray/tests/test_parametric.py
jasonb5: @dcherian See comment above, changed all the constants over and verified that's how they're stored in netCDF files.

cf_xarray/parametric.py
```diff
@@ -0,0 +1,821 @@
...
+        )
+
+        raise ValueError(
+            f"Could not derive standard name from combination of {stdnames}."
+        )
+
+    return found_stdname
+
+
+class ParamerticVerticalCoordinate(ABC):
```
dcherian, line 115:
```suggestion
class ParametricVerticalCoordinate(ABC):
```

PR #573, opened 2025-06-19
fix geometry detection

martinfleis:
Fixes https://github.com/xarray-contrib/xvec/issues/126

Neither the tests for the encoding in here and in xvec were able to catch this as we simply check the roundtrip, which does nothing in either direction...


martinfleis:
A tiny request - if we manage to merge and release this before Sunday's LPS Xvec workshop, I'd appreciate it. I can't even downgrade xarray as it breaks something else 😅

dcherian:
Thanks!

PR #582, opened 2025-08-15
handle monotonicity of cftime objects

### LINKED ISSUE #581: `v0.10.7` breaks `bounds_to_vertices` for cftime objects
I just realized my CI broke because, e.g., this does not work since the latest release (worked before):

```python
# Example: xarray Dataset with CF-compliant time bounds using cftime
import numpy as np
import xarray as xr
import cf_xarray as cfxr
import cftime

# Create cftime objects for monthly bounds
periods = 3
start = cftime.DatetimeGregorian(2000, 1, 1)
edges = [cftime.DatetimeGregorian(2000, m, 1) for m in range(1, periods + 2)]

# Bounds as [start, end) for each month
bnds = np.array([[edges[i], edges[i+1]] for i in range(periods)])
mid = np.array([edges[i] + (edges[i+1] - edges[i]) / 2 for i in range(periods)])

# Sample data
values = xr.DataArray(np.arange(periods, dtype=float), dims=("time",), coords={"time": mid})

# Build dataset with CF-style bounds
ds = xr.Dataset(
    {"foo": values},
    coords={
        "time": ("time", mid, {"bounds": "time_bnds"}),
        "time_bnds": (("time", "bnds"), bnds),
        "bnds": ("bnds", [0, 1])
    }
)

# Create vertices from bounds
cfxr.bounds_to_vertices(ds.time_bnds, "bnds")
```
gives...
```
TypeError                                 Traceback (most recent call last)
Cell In[3], [line 30](vscode-notebook-cell:?execution_count=3&line=30)
     20 ds = xr.Dataset(
     21     {"foo": values},
     22     coords={
   (...)     26     }
     27 )
     29 # Create vertices from bounds
---> 30 cfxr.bounds_to_vertices(ds.time_bnds, "bnds")

File ~/python/packages/cf-xarray/cf_xarray/helpers.py:181, in bounds_to_vertices(bounds, bounds_dim, core_dims, order)
    174     raise ValueError(
    175         f"Bounds format not understood. Got {bounds.dims} with shape {bounds.shape}."
    176     )
    178 core_dim_coords = {
    179     dim: bounds.coords[dim].values for dim in core_dims if dim in bounds.coords
    180 }
--> 181 core_dim_orders = _get_core_dim_orders(core_dim_coords)
    183 return xr.apply_ufunc(
    184     _bounds_helper,
    185     bounds,
   (...)    196     output_dtypes=[bounds.dtype],
    197 )

File ~/python/packages/cf-xarray/cf_xarray/helpers.py:240, in _get_core_dim_orders(core_dim_coords)
    237 if nonzero_diffs.size == 0:
    238     # All values are equal, treat as ascending
    239     core_dim_orders[dim] = "ascending"
--> 240 elif np.all(nonzero_diffs > 0):
    241     core_dim_orders[dim] = "ascending"
    242 elif np.all(nonzero_diffs < 0):

TypeError: '>' not supported between instances of 'datetime.timedelta' and 'int'
```
It still, however, works with usual datetime objects.

larsbuntemeyer:
closes #581

Improves the handling of datetime timedelta objects in core dimension ordering and bounds monotonicity checks within the `cf_xarray/helpers.py` module. The changes ensure that comparisons involving timedeltas are correctly performed using their total seconds.

Enhancements for datetime timedelta support:

- [x] Updated `_get_core_dim_orders` to detect and process `datetime.timedelta` differences using the `total_seconds` method for safe numeric comparison.
- [x] Refactored `_is_bounds_monotonic` to check for `datetime.timedelta` in bounds differences and convert them to seconds before monotonicity evaluation, instead of always casting to float64.
- [x]  Add tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,7 +37,7 @@ jobs:
             python-version: "3.13"
             os: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           fetch-depth: 0 # Fetch all history for all branches and tags.
       - name: Set environment variables
@@ -70,7 +70,7 @@ jobs:
         run: |
           pytest -n auto --cov=./ --cov-report=xml
       - name: Upload code coverage to Codecov
-        uses: codecov/codecov-action@v5.4.3
+        uses: codecov/codecov-action@v5.5.0
         with:
           file: ./coverage.xml
           flags: unittests
@@ -88,7 +88,7 @@ jobs:
       matrix:
         python-version: ["3.11", "3.13"]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           fetch-depth: 0 # Fetch all history for all branches and tags.
       - name: Set up conda environment
@@ -109,7 +109,7 @@ jobs:
         run: |
           python -m mypy --install-types --non-interactive --cobertura-xml-report mypy_report cf_xarray/
       - name: Upload mypy coverage to Codecov
-        uses: codecov/codecov-action@v5.4.3
+        uses: codecov/codecov-action@v5.5.0
         with:
           file: mypy_report/cobertura.xml
           flags: mypy
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'xarray-contrib/cf-xarray'
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           fetch-depth: 0
       - uses: actions/setup-python@v5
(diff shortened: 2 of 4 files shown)


aulemahal (commented):
Looks good! Thanks for the fix, and sorry for the breaking change. I didn't think of time dtypes when reviewing this.

Out of curiosity, what workflow are you doing that requires a `bounds_to_vertices` on the time bounds ?

I can approve and release the patch when a test is added.

larsbuntemeyer:
> Out of curiosity, what workflow are you doing that requires a bounds_to_vertices on the time bounds ?

I do lots of cmorization! Sometimes need it for the [cmor.axis](https://github.com/PCMDI/cmor3_documentation/blob/main/mydoc/mydoc_cmor3_api.md#cmor_axis) API, but now as you asked, i looked closer into the `cell_bounds` keyword and it seems it also accepts the 2d version with (2, n) dims.

dcherian:
@aulemahal can you take a look please?

dcherian (approved):

PR #595, opened 2025-09-18
Raise ValueError in `_get_ordered_vertices()` with "mixed" order

### LINKED ISSUE #594: Variable endpoints undefined when order is not ascending or descending
Today I ran into this error:

> UnboundLocalError: cannot access local variable 'endpoints' where it is not associated with a value

Sorry I don't have a minimal reproducible example yet. I'll try to make one ASAP.
Just posting in case others already know what's going on.

From what I can tell, this happens because the case where `order not in ["ascending", "descending"]` isn't covered.
See: https://github.com/xarray-contrib/cf-xarray/blob/70be853968bb90da13cf4f44b5399b64f4f5a93e/cf_xarray/helpers.py#L335-L340

This seems to have been introduced with PR https://github.com/xarray-contrib/cf-xarray/pull/579
Maybe @tomvothecoder has thoughts?

tomvothecoder:
A user reported an issue in #594 where `_get_ordered_vertices` fails to set the `endpoints` variable when the order of core dims are `"mixed"`. In their case, the longitude coordinates are circular (0–360), and the 1-D core dimension isn’t strictly monotonic because it crosses the seam at 0°.

We should raise a `ValueError` for `"mixed"` cases, rather than silently normalize the coordinates. 

- Closes #594 

## Long-term solution
Longer term, somebody can open a PR to consider opt-in support for circular axes, e.g. `bounds_to_vertices(..., circular_period=360.0, start=None)`. With `circular_period` set, cf-xarray could detect circular monotonicity, rotate away from the seam, and proceed safely.

### DIFF
--- a/cf_xarray/helpers.py
+++ b/cf_xarray/helpers.py
@@ -338,6 +338,12 @@ def _get_ordered_vertices(
         elif order == "descending":
             endpoints = np.maximum(bounds[..., :, 0], bounds[..., :, 1])
             last_endpoint = np.minimum(bounds[..., -1, 0], bounds[..., -1, 1])
+        elif order == "mixed":
+            raise ValueError(
+                "Cannot determine vertices for non-monotonic bounds with mixed core "
+                "dimension orders. Try normalizing the coordinates to a monotonic "
+                "convention and try again."
+            )
 
         vertices = np.concatenate(
             [endpoints, np.expand_dims(last_endpoint, axis=-1)], axis=-1


cf_xarray/helpers.py
```diff
@@ -338,6 +338,12 @@ def _get_ordered_vertices(
         elif order == "descending":
             endpoints = np.maximum(bounds[..., :, 0], bounds[..., :, 1])
             last_endpoint = np.minimum(bounds[..., -1, 0], bounds[..., -1, 1])
+        elif order == "mixed":
```
malmans2, line 341: Even if this is a private function, I would avoid having code that can result in an `UnboundLocalError`.
I suggest either replacing it with a bare else, or adding something like:
```python
else:
    raise NotImplementedError(f"{order = }")
```

tomvothecoder: The new code raises a `ValueError` so it shouldn't result in `UnboundLocalError`. However, I updated this conditional to `else` as a fall-back for any `order` value (even if it should only be ascending, descending, or mixed).

PR #622, opened 2026-03-25
Avoid overlapping bounds and/or gaps when points are irregularly spaced

bouweandela:
The current implementation of `cf.add_bounds` may introduce overlapping bounds or bounds with gaps in between when the points are irregularly spaced. This pull request changes the algorithm for guessing the bounds so the bounds are always in the middle of two points. The lower bound of the first point and the upper bound of the last point are extrapolated as before. This is similar to how this feature has been implemented in `iris`: https://github.com/SciTools/iris/blob/0fa0d6bfd0446ee5cd52c0ac6372e6bef5b250ec/lib/iris/coords.py#L2470-L2477

It also fixes an issue where the dimensions of the bounds variable are transposed w.r.t. the main variable.


aulemahal (approved):
This looks good! Thanks!

bouweandela:
Thank you for the quick review! It looks like there is an issue with coverage reporting: #623.

cf_xarray/helpers.py
```diff
@@ -23,15 +23,16 @@ def _guess_bounds_1d(da, dim):
...
-    upper = da + diff / 2
-    bounds = xr.concat([lower, upper], dim="bounds")
+    bound_position = 0.5
+    diff = da.diff(dim).pad({dim: (1, 1)}, mode="edge").reset_index(dim)
+    lower = (
+        da.reset_index(dim) - bound_position * diff.isel({dim: slice(0, -1)})
+    ).assign_coords({dim: da[dim]})
+    upper = (
+        da.reset_index(dim) + bound_position * diff.isel({dim: slice(1, None)})
+    ).assign_coords({dim: da[dim]})
```
dcherian, line 33: oops, merged too early.

we could optimize here by calling `reset_index` once and assigning coords to `result` directly

bouweandela: No worries, I'm happy to continue working on this. I'm still pretty new to xarray, but I think I found a better way to do this and opened a pull request in #627.

PR #630, opened 2026-04-01
Fix mypy checks

bouweandela:


cf_xarray/accessor.py
```diff
@@ -1814,10 +1814,6 @@ def _rewrite_values(
                         for k, v in value.items()
                     )
                 )
-
-            elif value is Ellipsis:
```
bouweandela, line 1818: This code is unreachable because the function `always_iterable` never returns `Ellipsis`

